PR #37, opened 2015-07-31
Support for (remote) object stores

ktf:
This introduces a new option `--remote-store <url>` which can be used to recycle already built packages, by picking up their binary tarballs from a common object store.


dberzano:
Hello, before leaving for my short absence here are my comments.

* f062f20: this is great, we badly needed this :+1:

* `--remote-store` works, I have tried it with the `::rw` option too, however package upload only occurs after building them. That is, if I build all without `--remote-store`, packages are built locally. If I re-run it with the option, packages are not uploaded. For consistency, I'd add it :-)

ktf:
I suggest we merge this if you're happy and I will fix the second comment
once I am back.

dberzano:
Sure, agreed :+1:

PR #50, opened 2015-08-25
Add support for incremental builds.

ktf:
When developing locally using the `--devel` mode, if the external is well
behaved and supports incremental building, it is possible to specify an
`incremental_recipe` in the YAML preamble. Such a recipe will be used after the
second time the build happens (to ensure that the non incremental parts of the
build are done) and will be executed directly in $BUILDDIR, only recompiled
what changed. Notice that if this is the case the incremental recipe will
always be executed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,3 +208,13 @@ and it will correctly install everything in `sw/<arch>/ROOT/latest`.
 
 It's also important to notice that if you use the devel mode, you will not be
 able to write to any store and the generated tgz will be empty.
+
+### Incremental builds
+
+When developing locally using the `--devel` mode, if the external is well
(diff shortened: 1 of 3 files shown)


dberzano:
:+1: works for me, tested with alisw/alidist#61, *i.e.* ROOT, AliRoot and AliPhysics. Not merging, leaving it to you (@ktf).

Wondering if it's possible (and does not clutter code too much) to have some kind of `--devel-recipe` to avoid recompiling all when changing the recipe itself? (Just leaving this as a note, TBD separately.)

ktf:
Ok, merging this.

I'm not sure I understand what you mean. Can't you just say `build <something-to-develop>`?

PR #63, opened 2015-09-07
Sort dependencies using a proper topological sort.

### LINKED ISSUE #60: Direct dependency cannot be a direct dependency as well.
Looks like if we have somethink like this in `root.sh`:

```
...
requires:
  - Alien-runtime
  - CMake
...
```

and:

```
...
requires:
  - CMake
...
```

in Alien-runtime we end up with:

```
DEBUG:AliRoot:63655bd0: Calculating hashes.
  File "alibuild/aliBuild", line 301, in <module>
    h.update(specs[dep]["hash"])
KeyError: 'hash'
```

this is due to some wrong dependency ordering.

ktf:
Previous algorithm was not taking into account more than 1 level of dependencies. This should fix #60.

### DIFF
--- a/aliBuild
+++ b/aliBuild
@@ -174,8 +174,6 @@ if __name__ == "__main__":
   while packages:
     p = packages.pop(0)
     if p in specs:
-      buildOrder.remove(p)
-      buildOrder.insert(0, p)
       continue
     try:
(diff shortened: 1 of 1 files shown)


ktf:
@dberzano this seems to work for me. Can you try it out?

dberzano:
Trying it now. Give me a sec :-)

PR #77, opened 2015-09-25
Allow build to happen inside a docker container

ktf:
This is useful for cross platform builds. No more having to spawn the
containers on your own. For the moment only slc7 is used, but it's easy
to allow any kind of container.


ktf:
@dberzano I bet you are gonna love this. ;)

Issue #61, opened 2015-09-03
Wrong specfile picked by aliBuild under certain conditions?


dberzano:
See for instance [Build #988](https://alijenkins.cern.ch/job/build-any-ib/988/consoleText).

So, the GSL recipe was changed in alisw/alidist#78 to get the sources from Git (using the `sources:` parameter) as opposed to downloading the tarball (in the recipe's body).

What we can see from the following log is that:

 * aliBuild correctly clones the new Git repository
 * but it runs the old recipe's body from /build/workarea/sw/20150829/SPECS/GSL.sh

Here's the log:

```
DEBUG:AliPhysics:vAN-2015: + mkdir -p /build/workarea/sw/20150829/INSTALLROOT/bbde5365b7427b311b02f8ea4d3b7576ac3a81c8/slc5_x86-64/GSL/v1.16-1 /build/workarea/sw/20150829/BUILD/bbde5365b7427b311b02f8ea4d3b7576ac3a81c8 /build/workarea/sw/20150829/BUILD/bbde5365b7427b311b02f8ea4d3b7576ac3a81c8/GSL /build/workarea/sw/20150829/INSTALLROOT/bbde5365b7427b311b02f8ea4d3b7576ac3a81c8/slc5_x86-64/GSL/v1.16-1
DEBUG:AliPhysics:vAN-2015: + cd /build/workarea/sw/20150829/BUILD/bbde5365b7427b311b02f8ea4d3b7576ac3a81c8
DEBUG:AliPhysics:vAN-2015: + ln -snf bbde5365b7427b311b02f8ea4d3b7576ac3a81c8 /build/workarea/sw/20150829/BUILD/GSL-latest
DEBUG:AliPhysics:vAN-2015: + [[ ! git://git.savannah.gnu.org/gsl.git == '' ]]
DEBUG:AliPhysics:vAN-2015: + [[ g != \/ ]]
DEBUG:AliPhysics:vAN-2015: + [[ ! -d /build/workarea/sw/20150829/SOURCES/GSL/v1.16/release-1-16 ]]
DEBUG:AliPhysics:vAN-2015: + rm -rf /build/workarea/sw/20150829/SOURCES/GSL/v1.16/release-1-16
DEBUG:AliPhysics:vAN-2015: + git clone git://git.savannah.gnu.org/gsl.git /build/workarea/sw/20150829/SOURCES/GSL/v1.16/release-1-16
[...]

PR #78, opened 2015-09-28
Jenkins script for steering builds

dberzano:
@ktf this is a request for comments: I think we should put the script we use in Jenkins to trigger build-any-ib under version control.

This script uses a separate build directory for each Mesos worker. Apparently the build problems we've kept encountering with autotools (see alisw/alidist#112) are gone with clean working directories...

Note that the script is sub-optimal as it deletes the working directory at the end of its job, so you can't examine the situation if something goes wrong.

### DIFF
--- a/aliJenkins
+++ b/aliJenkins
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+MIRROR=/build/mirror
+WORKAREA=/build/workarea/$NODE_NAME
+
+echo "=== ENVIRONMENT ==="
+env
+echo "=== /ENVIRONMENT ==="
+echo "Host: $(hostname)"
+lsb_release -a 2>/dev/null
+echo "uname: $(uname -a)"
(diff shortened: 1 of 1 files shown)


ktf:
What do you mean by "working directories"? Please hold on any change. This potentially leaves unhandled areas behind.

dberzano:
Yes indeed, this PR was made for showing an example and must not be merged as is. For reference, we have the same build started with the same parameter:

 * [this](https://alijenkins.cern.ch/job/build-any-ib/1533/) was done with no clean workarea, and it failed
 * [this](https://alijenkins.cern.ch/job/build-any-ib-test/14/) was done with the above script just to ensure we have a clean workarea, and it succeeded

I am closing this showcase and open a new issue instead.

dberzano:
Note that this problem seems to be #61.

PR #81, opened 2015-10-05
Publish requires which are runtime only

ktf:
When a remote store is specified, we also populate the toplevel
directory "dist-runtime" which contains only the packages which are
required at runtime. This is done so that we can avoid publishing things
like autotools and cmake on Alien.


dberzano:
:+1: great!

ktf:
Let's try this in next for a couple of builds and then we can rebase next and merge it to master, including the docker support. Notice that at the moment if something it's a build_requires, all its dependencies will have to be build_requires, otherwise they will be picked as runtime ones.

dberzano:
Which makes perfect sense to me.

PR #102, opened 2015-12-04
Disable shell path caching

dberzano:
Bash caches executables in $PATH for faster retrieval: we disable it to prevent
intermittent problems when we need to use a just-built package that, for
instance, overrides a system tool already in the cache. The latter could get
invoked instead with caching on (which is the default in Bash).


dberzano:
@ktf: eventually this did not fix the problem we had, but it is pretty harmless and can turn out useful... can we merge it?

PR #132, opened 2016-01-27
Use shorter logic for detecting the number of cpus

ktf:


ktf:
@dberzano: any reason we shouldn't shorten it to this?

dberzano:
@ktf: SC_NPROCESSORS_ONLN might not exist and an exception is thrown. What was committed works and was thought through, honestly I would not waste any more second on this :-)

ktf:
What committed will still have to be maintained. Less lines, less longterm maintenance. ;-)

Is there any supported platform where `SC_NPROCESSORS_ONLN` is not there? I could not find any.

PR #133, opened 2016-01-27
Cleanup unneeded build by products

ktf:
Cleanup build byproducts after the build. In particular it removes
INSTALLROOT/<package-hash> and BUILD/<package-hash> which are not really
needed in case of a successful build.

Notice this only happens in case a package is not a development package
since in that case build byproducts will be reused in a subsequent
build.


ktf:
I thought I merged this in #128 . Not sure what happened.

PR #137, opened 2016-03-09
Minor fixes for weird characters in logs and RHEL auto-detect.

ktf:


ktf:
@Barthelemy this should take care of your issue with RHEL6.

Barthelemy:
To answer your question in the email, this is not RHEL6 but SLC6. I get the same output on lxplus : 

> -bash-4.1$ python -c "import platform ; print platform.dist()" 
> ('redhat', '6.7', 'Carbon')

ktf:
I understand. It's the docker images we use for the builds which are actually centos . Anyways, can you confirm this fixes the issue?

Barthelemy:
I will try it tomorrow.

Barthelemy:
It works.

PR #143, opened 2016-03-21
Fix Unicode decode error in Riemann

dberzano:


dberzano:
@ktf: this should do the trick

Issue #144, opened 2016-03-22
alibuild fails on Mac if sha1sum is not installed


Barthelemy:
When sha1sum is not installed on Mac, aliBuild fails with this error 

> Unable to detect source code changes.

Maybe you could detect it and provide a more insightful message.

ktf:
Actually I think I should simply use `shasum` in place of `sha1sum`, which exists on mac.

Barthelemy:
Agreed.

ktf:
Closing this as I think it's fixed in `next` branch and will go in `master` soon. Reopen if not the case.

PR #146, opened 2016-03-22
Use shasum -a 1 rather then sha1sum

ktf:


ktf:
@barthelemy can you check this? It should fix alisw/alibuild#144.

PR #147, opened 2016-03-23
Improve aliModules and simplify its usage

dberzano:
More intuitive usage with sensible defaults: it works out of the box with zero
parameters.

Also detects modulecmd on various platforms, including Homebrew installations on
OSX. A platform-specific installation hint is printed if modulecmd is missing.


dberzano:
As discussed with @ktf, further improvements to converge with CVMFS's `alienv` are on the way. This is to guarantee a more uniform user experience between the aliBuild and CVMFS cases.

Issue #142, opened 2016-03-18
List available defaults


ktf:
It would be great to have a list of available defaults if a wrong one is provided.

PR #150, opened 2016-03-23
List available defaults when a wrong one is specified

ktf:
Also print out the default if `--defaults` is specified without an
option.


ktf:
Fix #142 .

PR #155, opened 2016-03-24
Detect current shell type in alienv

dberzano:


dberzano:
Force-pushed by mistake while opening #156. Re-opened as #157 for clarity.

PR #156, opened 2016-03-27
Correctly relocate modulefile in devel mode

dberzano:
Same rationale as #153 but:

 - works if recipe does not provide a modulefile
 - works with all sed flavors (i.e. on Ubuntu, `-i` with extension is not
   supported)


ktf:
Yes, I was testing a similar fix

PR #157, opened 2016-03-27
Detect current shell type in alienv

dberzano:


dberzano:
Note: still suffers from a potential environment problem in `alienv enter` in case user has some garbage in `.bashrc`. Leaving it open for reference, but there is still some work to do.

PR #159, opened 2016-03-27
Store references by default as a subdir of workdir

dberzano:
Rationale: if running `aliBuild -w yabbayabba -d build YabbaYabba` I expected the reference sources to be under `yabbayabba/MIRROR` but they were not stored anywhere - it attempted to store them under `sw/MIRROR`, but `sw` did not exist and it gave up. This PR defaults reference sources as the `MIRROR` subdirectory of the current working directory whatever it is called.


dberzano:
@ktf, for your consideration.

PR #162, opened 2016-03-28
Do not clean install root for devel packages

dberzano:
@ktf: incrementally building packages has the side effect of removing modulefiles, as they are created in the full recipe.

Not sure about this solution; this speeds up the incremental builds, however it might introduce inconsistencies too, _e.g._ when a file/library is removed in the source and it will still exist in the destination.

TBH I think that:

 - recipes should be changed instead, to put the modulefile to the install directory also in the incremental versions
 - if users want to speed up things, instead of using `aliBuild` (which always guarantees consistency) they should `cd $BUILDDIR && make install` manually

In practice I would **not** merge this, but you might find a reason why this is the way to go instead?

### DIFF
--- a/aliBuild
+++ b/aliBuild
@@ -853,7 +853,8 @@ if __name__ == "__main__":
     if fileHash != spec["hash"]:
       if fileHash != "0":
         debug("Mismatch between local area and the one which I should build. Redoing.")
-      shutil.rmtree(dirname(hashFile), True)
+      if not spec["package"] in develPkgs:
+        shutil.rmtree(dirname(hashFile), True)
     else:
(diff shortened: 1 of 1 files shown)


ktf:
I fully agree on point n 2. At least for the moment, aliBuild should be used to setup the development environment, not as a build tool.

Regarding 1, I think we should have automatic modulefile creation in alibuild, with some way to overload special cases. That would open the door to proper behaviour also in incremental builds.

dberzano:
@ktf yup, agreed. Recipes have been modified accordingly for now (see PR). This can be closed.

PR #177, opened 2016-04-06
aliBuild prepare

dberzano:


dberzano:
@ktf let's discuss/amend it first

aliBuild
```diff
@@ -422,6 +424,11 @@ if __name__ == "__main__":
                       dest="autoCleanup", action="store_false", default=True)
   parser.add_argument("--devel-prefix", "-z", nargs="?", help="Version name to use for development packages. Defaults to branch name.",
                       dest="develPrefix", default=argparse.SUPPRESS)
+
+  parser.add_argument("--devel-packages", "-p", dest="develPackages",
```
ktf: Why not using "--devel"?

aliBuild
```diff
@@ -114,8 +114,10 @@ def updateReferenceRepos(referenceSources, p, spec):
   # If the area can be created, clone a bare repository with the sources.
   debug("Updating references.")
   referenceRepo = "%s/%s" % (abspath(referenceSources), p.lower())
-  if os.access(dirname(referenceSources), os.W_OK):
-    getstatusoutput("mkdir -p %s" % referenceSources)
+  try:
```
ktf: Why do you need to change this?

ktf:
yes, indeed. We need to brainstorm a bit what the UI looks like. Can you provide an example workflow of how you plan to have this to be used?

dberzano:
Example workflow. For multiple directories:

```
aliBuild prepare --devel-packages AliRoot,AliPhysics [--alidist IB/v5-08/next] alice-master
cd alice-master/
aliBuild build AliPhysics -d -w ../sw -z
```

This fetches clones in the `alice-master/` directory, by using reference repos (so to save space).

Note that it also works for people that prefer to have a single directory:

```
aliBuild prepare --devel-packages AliRoot,AliPhysics [--alidist IB/v5-08/next] .
aliBuild build AliPhysics -d
```

aliBuild
dberzano: Yes no problem.

dberzano:
So this removes all custom switches and reuses existing ones for `aliBuild build`. One can now do:

```
aliBuild prepare zlib
```

for downloading `zlib` for development mode in the current directory. Example with AliRoot/AliPhysics in a custom subdirectory:

```
aliBuild prepare AliRoot,AliPhysics -z my-aliroot-set [--dist IB/v5-08/next]
```

The `--dist` switch has been retained and it is honored in `aliBuild build` too.

Moreover, `--config-dir` is honored too.

The reference sources tree creation has been moved out of `updateReferenceRepos` as discussed.

To be discussed tomorrow :-)

ktf:
Yes, I like this one much more. Maybe we could also introduce a syntax to checkout a given tag / branch:

```
aliBuild prepare AliRoot@v5-08-01,AliPhysics@v5-08-01-1 -c ktf/alidist@IB/v5-08/next
```

On a separate note, maybe we should also consider renaming of alibuild to alisw:

```
alisw prepare ...
alisw build ...
alisw env ...
```

ktf:
Of course not for this PR.

dberzano:
OK so now we have `init` instead of `prepare`.

As for the rest, you can already check out a version, with the syntax:

```
aliBuild prepare AliRoot=feature-onlineqa,AliPhysics=master
```

If not provided, version defaults to the spec's `tag`. Existing directories are skipped for the moment: if e.g. AliRoot exists already, it does _not_ checkout the correct version but it skips it. As discussed, I'd keep it like this with a warning, and (for a later PR) I'd add a `--force` to forcibly checkout the desired version and erase user's content.

As for the `alisw <whatever>` it seems a good idea to me.

PR #181, opened 2016-04-08
Fix for when log message is not a string

dberzano:
This is actually a bug that needs to be fixed quickly :-(


ktf:
From now on we wait for **at least** one week before merging anything into production... ;)

PR #216, opened 2016-04-21
Change default interpreter from python to python2

akubera:
If the user is using python3 as their default python interpreter, running aliBuild or aliDoctor causes a SyntaxError. Until universal python support is implemented, changing the 'python' executable to 'python2', as perscribed in [PEP 394](https://www.python.org/dev/peps/pep-0394/), will select the right version.

### DIFF
--- a/aliBuild
+++ b/aliBuild
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 import os, logging, time, re, yaml, hashlib, argparse
 import sys, shutil, subprocess, socket
 from commands import getstatusoutput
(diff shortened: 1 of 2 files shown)


akubera:
Actually, it may be the case that the 'python2' symlink is not in OSX's default PATH. I believe "python2.7" would work universally.

Alternatively it might not be too difficult to make the scripts python3 compatible.

ktf:
The problem is that python 2.7 is not there everywhere (not on SLC5 which is still our production architecture). I personally think making it compatible with python3 is probably the best option. I gave a look with 2to3 and indeed there is only minor changes and a few modules for which we would need dual support, but apart from that is not impossible. If you feel like giving it a stab, I would also appreciate.

PR #219, opened 2016-04-25
Add LIBENVNAME and SONAME variables

ktf:
On mac:

    LIBENVNAME=DYLD_LIBRARY_PATH
    SONAME=dylib

On linux:

    LIBENVNAME=LD_LIBRARY_PATH
    SONAME=so


ktf:
We decided to go with an alternative route. Closing this.

PR #222, opened 2016-04-27
Improve architecture detection

ktf:
Now the helper for architecture detection is pure and has its own
unittest. This should make sure we do not inadvertly change the
architecture while trying to support a different one.


ktf:
@qgp the change I want to do is this one, so keep it into account when you write your Gentoo support.

PR #225, opened 2016-04-27
Improve packaging

ktf:
- Add argsparse as a dependency (so that it works out of the box on
  certain python 2.6 installations.
- Distribute shell helpers.


ktf:
Nothing actually checks this, right now. Skipping tests.

PR #240, opened 2016-05-03
Remove README.md and fix syntax in README.rst

akubera:
PyPI correctly renders ReStructuredText files, so let's just use that as readme, instead of duplicating information.


ktf:
Indeed having one single README is probably a good idea, and since PyPI apparently does not support markdown, let's simply use rst, as you suggest. Thanks.

PR #250, opened 2016-05-04
Regexp special character \? not working on all systems. Replaced with \{0,1\}

aphecetche:
On my system where the command in the ps results is "-zsh" the \? of
the regular expression used to get the shell name do not seem to work fine...


dberzano:
Hello @aphecetche, I had got something similar with #229. Thanks for the fix, merging it!

PR #253, opened 2016-05-07
Make build dir for devel packages depend on prefix

### LINKED ISSUE #212: Different development branches of same software not handled correctly
Hi all,
I've installed two different development versions of aliroot.
For one, I just use the master:
aliBuild init -w ../sw AliRoot@master,AliPhysics -z ali-master --dist IB/v5-08/next

and for the other, I need a specific branch:
aliBuild init -w ../sw AliRoot -z ali-muonhlt --dist IB/v5-08/next
cd ali-muonhlt/AliRoot
git checkout feature-muonhlt

When I compile:
in ali-master: aliBuild -z -w ../../sw --disable GEANT4,GEANT4_VMC build AliPhysics
and 
in ali-muonhlt: aliBuild -z -w ../../sw --disable GEANT4,GEANT4_VMC build AliRoot

I find this in the BUILD directory:
AliRoot-latest -> d20faaa6662993d882eef0df002b58f08e8754d0
AliRoot-latest-ali-master -> d20faaa6662993d882eef0df002b58f08e8754d0
AliRoot-latest-ali-muonhlt -> d20faaa6662993d882eef0df002b58f08e8754d0

i.e., despite the AliRoot source code is different (I have the AliRoot master in ali-master and the branch feature-muonhlt in ali-muonhlt), the two different builds are assigned the same hashtag...and each time I compile one of the two, I automatically destroy the other.

Is there some options that I'm missing which would provide a clearly separate build for two clearly different development versions?

Thanks in advance,
best regards,
Diego

dberzano:
Fixes #212.

With this fix when using a `develPrefix` this is used to compute the hash to determine the build directory. This means that the same package, in development mode in two different prefixes, has two distinct build (and install) dirs.

### DIFF
--- a/aliBuild
+++ b/aliBuild
@@ -840,12 +840,15 @@ if __name__ == "__main__":
       h(specs[dep]["hash"])
     if bool(spec.get("force_rebuild", False)):
       h(str(time.time()))
-    if spec["package"] in develPkgs and "incremental_recipe" in spec:
-      h(spec["incremental_recipe"])
-      incremental_hash = hashlib.sha1(spec["incremental_recipe"]).hexdigest()
-      spec["incremental_hash"] = "INCREMENTAL_BUILD_HASH=%s" % incremental_hash
-    elif p in develPkgs:
-      h(spec.get("devel_hash"))
+    if p in develPkgs:
+      if "incremental_recipe" in spec:
+        h(spec["incremental_recipe"])
+        incremental_hash = hashlib.sha1(spec["incremental_recipe"]).hexdigest()
+        spec["incremental_hash"] = "INCREMENTAL_BUILD_HASH=%s" % incremental_hash
+      else:
+        h(spec.get("devel_hash"))
+      if "develPrefix" in args:
+        h(args.develPrefix)
     spec["hash"] = h.hexdigest()
     debug("Hash for recipe %s is %s" % (p, spec["hash"]))
 
--- a/aliBuild
+++ b/aliBuild
@@ -828,24 +828,24 @@ if __name__ == "__main__":
   debug("Calculating hashes.")
   for p in buildOrder:
     spec = specs[p]
-    h = hashlib.sha1()
+    h = Hasher()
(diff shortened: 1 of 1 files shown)


dberzano:
@dstocco this should fix #212. Merging as soon as possible after a brief test.

ktf:
There is actually something I do not understand of all this. A development package on two different commits should already end up in different places. The issue, if I understand correctly, is only for the case the packages have an incremental recipe. Am I wrong? If yes, why not keeping the code as it was and simply doing:

```
diff --git a/aliBuild b/aliBuild
index 3226cb7..4836ed4 100755
--- a/aliBuild
+++ b/aliBuild
@@ -843,6 +843,7 @@ if __name__ == "__main__":
       h.update(spec["incremental_recipe"])
       incremental_hash = hashlib.sha1(spec["incremental_recipe"]).hexdigest()
       spec["incremental_hash"] = "INCREMENTAL_BUILD_HASH=%s" % incremental_hash
+      h.update(args.develPrefix)
     elif p in develPkgs:
       h.update(spec.get("devel_hash"))
     spec["hash"] = h.hexdigest()
```

This should minimise side effects while still fixing #212.

dstocco:
Ciao @dberzano ,
I saw that now the BUILD directory is good: different local versions point to different build directories: good!

However, shouldn't it be the case also for the install dir?
Here is what I have in
sw/osx_x86-64/AliRoot/

lrwxr-xr-x   1 dstocco  staff     3 May  9 11:26 latest-ali-muonhlt -> 0-1
lrwxr-xr-x   1 dstocco  staff     3 May  9 12:43 latest-ali-master -> 0-1
lrwxr-xr-x   1 dstocco  staff     3 May  9 12:43 latest -> 0-1
drwxr-xr-x  49 dstocco  staff  1666 May  9 12:43 0-1

Thanks,
ciao,
Diego

dberzano:
@dstocco: indeed. Let us check & fix.

PR #277, opened 2016-05-19
all: explicitly use python2

### LINKED ISSUE #276: support for `python` as `python3`
Many Linux distributions made the switch to `python3`.

`aliBuild` and friends should explicitly request `python2` instead of `python` to cater for the future (or migrate to `python3` but that's understandably more work.)

sbinet:
Fixes #276.

### DIFF
--- a/aliBuild
+++ b/aliBuild
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 import os, logging, time, re, yaml, hashlib, argparse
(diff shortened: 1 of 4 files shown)


ktf:
python2 is not guaranteed to be there for ALICE supported architectures. I'd rather have something which works on both pytho2.6+ and python3. It should not be too complicated, but we have not got to it yet. PR (to "next" branch) are of course welcome.

PR #284, opened 2016-05-26
Add alibuild analytics subcommand

ktf:
This is used to programmatically turn on and off the sending of
analytics to Google Analytics.


dberzano:
OK apparently Travis **has** a tty :-)

We should touch the file to disable analytics in Travis first, elsewhere the test times out.

ktf:
Ciao,

I should have added `aliBuild analytics off`:

https://github.com/alisw/alibuild/pull/284/files#diff-354f30a63fb0907d4ad57269548329e3R21

which should have fixed the problem, so I think it’s actually 
something else.

--
Ciao,
Giulio

PR #292, opened 2016-05-30
Use commit_hash from local src for devel packages

dberzano:
Development packages used a random `commit_hash`, which came from a remote head (`git ls-remote`) and not necessarily for the currently checked out branch. This was not a problem until #282 where we use this piece of information to determine whether it is necessary to rebuild a development package.

If we get this information from the remote only, it might happen that package rebuild is not triggered in case we don't have local modifications, and in case we are moving back and forth pushed commits via `git reset --hard`.

### DIFF
--- a/aliBuild
+++ b/aliBuild
@@ -714,6 +714,11 @@ def doMain():
           # We are in development mode, we need to rebuild if the commit hash
           # is different and if there are extra changes on to.
           if spec["package"] in develPkgs:
+            # Devel package: we get the commit hash from the checked source, not from remote.
+            cmd = "cd %s && git rev-parse HEAD" % spec["source"]
+            err, out = getstatusoutput(cmd)
+            dieOnError(err, "Unable to detect current commit hash.")
+            spec["commit_hash"] = out.strip()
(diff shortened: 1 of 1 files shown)


dberzano:
@ktf merging it in `next` and doing rc2 for the moment - I have decided to keep the change minimal for clarity but the whole logic of the `for` loop might be refactored.

ktf:
Ok. Doesn't this force installation in a separate directory, thought?

ktf:
I.e. Shouldn't this be part of the devel_hash, rather than the commit_hash?

dberzano:
Hello, so I have checked and tested the following scenario:

* A new change has been published in a devel package: I do `git pull`
* I recompile: the same old build directory is used

So it should be safe. I have also checked the code:

* The place where we set `spec["commit_hash"]` for development packages also uses it right afterwards for computing other hashes, then it sets it to `0`.
* In order to calculate `spec["hash"]`, which is used for determining the `BUILD/<hash>` directory, `spec["commit_hash"]` is used, but at that point it is `0` already.

I'll wait for the official release in case other problems pop up. If you think of something overlooked by this PR please comment ;-)

PR #298, opened 2016-06-01
Feature python3 compat

ktf:


ktf:
I did a pull request for this to better discuss the issues. See inline comments.

ktf:
Needs rebase (and most likely a few more fixup ) after the merge of #295.

aliBuild
```diff
@@ -237,10 +247,10 @@ class HttpRemoteSync:
       else:
         hashList = json.loads(urlopen(hashListUrl).read())
         pkgList = json.loads(urlopen(pkgListUrl).read())
-    except URLError, e:
+    except URLError as e:
```
ktf: This is not compatible with python 2.4, which is currently used by slc5 (sigh). Can we factor this change out? We have in the plans to move to 2.6, but I still would like to do it separately.

alibuild_helpers/__init__.py
```diff
@@ -1,4 +1,7 @@
 # Dummy file to package build_template.sh
-import utilities as utilities
-import analytics as analytics
-import log as log
+
+from __future__ import absolute_import
+
+from . import utilities
```
ktf: Is this compatible with python on slc5? If not, can it be factored out in a separate PR?

akubera: It looks like `absolute_import` was introduced as a `__future__` in Python2.5 (see [PEP 328](https://www.python.org/dev/peps/pep-0328/)). That's unfortunate the lowest target is 2.4.

I might be able to put that in an version_checking if block. Does anyone know if `from . import X` is a syntax error in 2.4.

aliBuild
akubera: I think the only way to support the different syntaxes is to have separate files that are loaded by "globally compatible" (i.e. super simple/trivial) executable scripts (aliBuild, etc).

I think it would be possible to use the 2to3 script to minimize (human) code duplication. I don't know if it would be easy to apply to whole modules (such as the local yaml module).

ktf: You misunderstood me. I'd like the python 2.4 compatible changes to be in a separate PR from the python 2.6 only ones. This way we can easily revert the latter if we have problems with it.

akubera: Ah, I see; just to make merging easier down the line, not for simultaneous py3/py2.4 compatibility, which I don't think can happen easily.
Yeah I can create a new branch/pull request & cherry pick compatible changes. Should I also rebase this branch? I don't know if rebasing will break the pull request.

ktf: Correct. It's just to simplify merging. As I said I suspect you will have conflicts as I refactored a bit aliDoctor to improve user experience and to make it behave exactly like aliBuild

ktf:
Not needed anymore, I think. I have most of them handled in the new next branch.

ktf:
v1.4.0.rc1, which was just released, should work with python 3.

PR #308, opened 2016-06-06
Fix issue when setuptools is not there

ktf:


ktf:
Jenkins issue will be fixed in a separate PR.

ktf:
@kouzinopoulos if you pull from master your `pkg_resource` issue should be gone now.

PR #315, opened 2016-06-09
Silence harmless missing Toolchain error in alienv

### LINKED ISSUE #249: Silence the "no modulefile for Toolchain" message which is not an error

dberzano:
Fixes #249


ktf:
Shouldn't we merge this?

ktf:
(I think doing it in master is fine, since we do not use alienv in production)

PR #327, opened 2016-06-20
Protect again aliBuild in alienv cornercase

ktf:
aliBuild was not protecting against previously setup environment under
some circunstances. This should make sure the build environment will
always override the one previously set.

Notice there might still be leaks of the user environment into the build
one, but to have a definitive solution for this we would need to move to
rely on modulefiles only.

### DIFF
--- a/aliBuild
+++ b/aliBuild
@@ -307,6 +307,9 @@ def prunePaths(workDir):
       continue
     workDirEscaped = re.escape("%s" % workDir) + "[^:]*:?"
     os.environ[x] = re.sub(workDirEscaped, "", os.environ[x])
+  for x in list(os.environ.keys()):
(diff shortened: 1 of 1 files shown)


ktf:
@dberzano @dstocco this should fix the issue with running aliBuild from inside alienv. Can you please check?

dstocco:
Ciao Giulio,
I've pulled and recompiled O2, but unfortunately this does not fix issue [#451](https://github.com/alisw/alidist/issues/451).

However, I did not have the issues experienced by Chiara and Benjamin...so this might still fix their issue.

Ciao,
Diego

ktf:
One of the builders dies for unrelated reasons. The other tests worked fine. Merging.

PR #345, opened 2016-07-18
Default alienv to --no-refresh on read-only fs

dberzano:
In case of shared installations users using `alienv` do not normally have
permissions to regenerate the modulefiles directory. `alienv` now detects this
case and does not attempt to regenerate it. This saves users from adding
`--no-refresh` manually.


ktf:
Why not using `-w $directory`?

dberzano:
There are corner cases (reference pending, I will add one in case I find it - I recall something related to read-only filesystems, AFS, etc.) where a directory appears writable but it is not. The shortest reliable way to determine it is attempting to write something.

dberzano:
I am also afraid of some weird selinux/apparmor interplay reporting a directory is writeable when it's not. In any case Travis raises an error related to `alienv`, I am going to check it.

PR #349, opened 2016-08-01
Fix issue with wrong order of environment variables

fuhlig1:
Reading the content of the header section of any recipe results in an arbitrary order of the variables
defined in the env section. Often this order is importatnt and should be the same as defined in the
env section of the header.
Add functionality which loads the content of the header section into a ordered dictionary such that
the order of the variables is preserved.

### DIFF
--- a/aliBuild
+++ b/aliBuild
@@ -16,6 +16,10 @@ from alibuild_helpers.utilities import format, getVersion, detectArch
 from alibuild_helpers.analytics import decideAnalytics, askForAnalytics, report_screenview, report_exception, report_event
 from alibuild_helpers.analytics import enable_analytics, disable_analytics
 import traceback
+if float('%d.%d' % sys.version_info[:2]) < 2.7:
+  from ordereddict import OrderedDict
+else:
+  from collections import OrderedDict
 
 def writeAll(fn, txt):
   f = open(fn, "w")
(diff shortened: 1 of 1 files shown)


aliBuild
```diff
@@ -16,6 +16,10 @@ from alibuild_helpers.utilities import format, getVersion, detectArch
 from alibuild_helpers.analytics import decideAnalytics, askForAnalytics, report_screenview, report_exception, report_event
 from alibuild_helpers.analytics import enable_analytics, disable_analytics
 import traceback
+if float('%d.%d' % sys.version_info[:2]) < 2.7:
```
ktf: How about using:

```
try:
  from collections import OrderedDict
except ImportError:
  from ordereddict import OrderedDict
```

besides that `ordereddict` is not actually a standard python module, so, how about adding it also to the dependencies in `setup.py`?

fuhlig1: Hi Giulio

> try:
>   from collections import OrderedDict
> except ImportError:
>   from ordereddict import OrderedDict
> 
> besides that ordereddict is not actually a standard python module, so, how about adding it also to the dependencies in setup.py?

I am no Python expert so I simply hacked things together. I am not even sure if the try/except block is really needed. I found the block somewhere when looking
how to import the yaml stream into an ordered dictionary. 

Please fell free to change the commit as you like.

Ciao

Florian

ktf: Ok, fair enough. I will adapt it and make sure it works on slc5. Thank you for your contribution!

dberzano:
Hi, I am addressing the issue in #466, finally (and incidentally solving a build problem with Python 3). Closing this!

PR #352, opened 2016-08-15
Allow defaults to disable packages and overrides

ktf:
This should allow us to incorporate what is currently a different
branch to a set of overrides which can be specified in the defaults
file. The idea is that an `o2` default could for example disable
AliEn and switch to ROOT6.

### DIFF
--- a/aliBuild
+++ b/aliBuild
@@ -563,7 +563,6 @@ def doMain():
   workDir = abspath(args.workDir)
   prunePaths(workDir)
 
-
   if not exists(args.configDir):
     err = execute(format(
                     "git clone https://github.com/%(repo)s%(branch)s %(cd)s",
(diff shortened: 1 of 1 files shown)


ktf:
@MohammadAlTurany @fuhlig1 @Barthelemy @pzhristov @dberzano: this adds a feature which has been discussed in last CWG13 and which we discussed this morning with Mohammad  and Dario. 

It's now possible to specify "overrides" and "disables" in the defaults file, so that we can avoid a proliferation of branches but we can simply have small changes (e.g. compatible version changes) encoded in the different defaults files. For an example of how this can be used for o2, see also alisw/alidist#496 .

We can have a similar approach for Fair specific changes, e.g.:

```
aliBuild --defaults panda build panda
```

could be used to pick the versions required by panda.

ktf:
Jenkins issue is an unrelated out of memory when compiling with CMake. Merging.

PR #354, opened 2016-08-17
Fix system_requirements and docker

### LINKED ISSUE #339: When building in docker, prefer_system should check the container environment
When building in docker, we need to check the the system tools are available in the container itself, not in the host environment.

ktf:
This makes sure that system checks are done inside docker rather than
on the host. This should fix #339.


ktf:
Error unrelated, merging.

PR #355, opened 2016-08-17
Add --docker-image option

ktf:
This adds the ability to specify a docker image when building using
docker. If the option is not present, it will fall back to the usual,
architecture derived, one e.g. alisw/slc7-builder.


ktf:
The error is in slc7-daq-builder and it's an unrelated "out of memory" error when compiling boost. Merging.

PR #356, opened 2016-08-18
Allow spaces for the -e option

ktf:


ktf:
Trivial and does not affect production. Merging.

PR #359, opened 2016-08-18
Add recipes override from the defaults

dberzano:
Single recipes can be overridden (_e.g._ to provide a hotfix) in defaults using
overrides by specifying the package name followed by the Git reference (hash,
tag...) where the recipe should be picked from. Overrides are temporary and do
not alter the current checked out recipes directory.


dberzano:
@ktf Jenkins problem unrelated, merging...

Issue #325, opened 2016-06-17
alibuild clean: cannot handle multiple development packages


dstocco:
Dear all,
I used to have a development AliRoot/AliPhysics version linked to root v5:
ali-master:
  AliPhysics -> ../../refSrcs/AliPhysics/patches
  AliRoot -> ../../refSrcs/AliRoot/patches
  alidist(IB/v5-08/next)

I then decided to test root6. So I initialised a new alidist with IB/master/root6:

ali-root6:
  AliPhysics -> ../../refSrcs/AliPhysics/patches
  AliRoot -> ../../refSrcs/AliRoot/patches
  alidist(IB/master/root6)

My plan is to use both versions (ali-master and ali-root6), eventually slowly moving to root6.

However, when I did alibuild clean, it erased the root v5.
So I built ali-master again (it is fast since there is the TARS)...and this times it erased root v6.
In practice, it seems that the alibuild clean, cleans everything which is not associated to the "latest" build.
Can this be fixed so that, in case of several development versions, the clean only touches the packages that are used by nobody?

Thanks in advance,
best regards,
Diego

ktf:
Strange. This should be working just fine.

Are you using `aliBuild 1.3.0`? Can you setup both branches and then tell me what:

```
echo "Links"
find  sw/osx_x86-64/ -name "latest*"
echo "Clean"
aliBuild clean --dry-run
```

tells you?

dstocco:
Ciao Giulio,
I'm using aliBuild master from git (updated to yesterday).

In the meanwhile, I initialised a custom directory for ROOT6, with aliBuild init -z ali-root6 master
so that I can decide what version to use.
Anyways, I still have the problem with geant3, which is the only dependency left which changes between ali-root6 and ali-master.
Indeed, here comes your output:

Links

sw/osx_x86-64//AliEn-CAs/latest
sw/osx_x86-64//AliEn-Runtime/latest
sw/osx_x86-64//AliPhysics/latest
sw/osx_x86-64//AliPhysics/latest-ali-master
sw/osx_x86-64//AliPhysics/latest-ali-root6
sw/osx_x86-64//AliRoot/latest
sw/osx_x86-64//AliRoot/latest-ali-master
sw/osx_x86-64//AliRoot/latest-ali-muonhlt
sw/osx_x86-64//AliRoot/latest-ali-root6
sw/osx_x86-64//ApMon-CPP/latest
sw/osx_x86-64//boost/latest
[...]

PR #361, opened 2016-08-22
Introduce the concept of build family

ktf:
This introduces the concept of build family. Basically until now
only the development packages were getting a latest-<devel-prefix>
link, while any of their descendents did not. This leads to issue
when cleaning up as described in:

<https://github.com/alisw/alibuild/issues/325>

With this change we introduce a new concept, the build family, which is
common to all packages which are involved in the build of a development
package or which use a particular default. It allows us to create a
latest-<build-family> link for each package involved in the build.
<build-family> will be either:

- `<devel-prefix>-<defaults>`
- `<defaults>`

depending on whether the package is a devel package or not.

### DIFF
--- a/aliBuild
+++ b/aliBuild
@@ -1000,6 +1000,27 @@ def doMain():
     # - Use the latest number in the version, to decide its revision
     debug("Packages already built using this version\n%s" % "\n".join(packages))
     busyRevisions = []
+
+    # Calculate the build_family for the package
+    #
+    # If the package is a devel package, we need to associate it a devel
+    # prefix, either via the -z option or using its checked out branch. This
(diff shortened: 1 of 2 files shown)


ktf:
@dberzano please have a look at this. We can discuss tomorrow the details, but I think it should be fine.

ktf:
@dberzano @matthiasrichter This is the fix for the links I was referring too.

dberzano:
@ktf this makes sense to me

ktf:
**Should* be fixed.

ktf:
slc6, slc7 and slc5 build correctly. slc7-daq is lagging behind for unrelated issues. Merging.

PR #363, opened 2016-08-24
Self-heal setup after a full disk

ktf:
If aliBuild dies while creating the temporary archive, e.g. due to
a disk full, subsequent runs die trying to expand the half finished
tarball. While we have a protection which should move the tarball in
place atomically, the mechanism looking up for cached tarballs could
still pick up the .processing one. This makes sure all the tarball we
get actually end in `gz`, hence fixing the problem.

### DIFF
--- a/aliBuild
+++ b/aliBuild
@@ -1023,7 +1023,7 @@ def doMain():
 
     for d in packages:
       realPath = readlink(d)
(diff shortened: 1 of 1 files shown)


ktf:
Problem unrelated. Merging.

PR #365, opened 2016-08-29
Unify defaults / disable support in aliBuild / aliDoctor

ktf:
By using the same set of functions, aliBuild and aliDoctor should now
behave exactly the same. It's also now possible to pass --disable and
--defaults to aliDoctor and get the diagnosis for the chosen setup. By
factoring out the IO related code, this should also improve our ability
to unit test things.


ktf:
Error is unrelated. Still exercising this a bit.

ktf:
Errors are unrelated. Merging.

PR #371, opened 2016-09-13
Fix aliBuild clean

ktf:
Given a package aliBuild did not delete a given version, e.g. v1.0, if a
symlink called latest* existed in any package, not only in the one being
looked at. This caused a few packages not being deleted.

This resolves the issue by making sure that the `latest` link is
actually in the same directory.


ktf:
Unrelated build issue. Merging.

PR #376, opened 2016-10-19
Fix python3 incompatibilities

ktf:


ktf:
Error seems to be unrelated. Merging.

PR #394, opened 2016-11-29
Add tutorial for O2Group modules

Barthelemy:
As requested by @ktf, here is the little tutorial I made for O2 modules.


ktf:
Fine for now. Maybe we should add a few more usecases and have a separate "Example" folder?

Barthelemy:
I think that it is a good idea.

PR #398, opened 2016-12-07
Disable write store only at first devel package

ktf:
Previously write store was disabled in case a development package was
found in the dependency tree. This changes the behavior to be less
aggressive and disables the ability to write to the store only once the
first development package is encountered while building.


ktf:
Works in Aurora, merging.

PR #408, opened 2017-01-11
Introduce valid_defaults metadata

ktf:
Some packages can only be compiled using a given default. Most notably,
the O2 package can only be built with `--defaults o2` or `--defaults
o2-daq`. It's now possible to express that by adding a valid_defaults
entry in the YAML preable which contains the list of valid defaults.


ktf:
Jenkins issue unrelated.

PR #415, opened 2017-01-24
Apply defaults only to overriden packages

ktf:
Before this patch, different defaults were effectively generating
different packages. This makes sure that only the bits which have
been overridden trigger a rebuild of the overriden package and its
dependencies, but not more.

### DIFF
--- a/aliBuild
+++ b/aliBuild
@@ -651,7 +651,7 @@ def doMain():
 
   if buildOrder:
     banner("Packages will be built in the following order:\n - %s" %
-           "\n - ".join([ x+" (development package)" if x in develPkgs else "%s@%s" % (x, specs[x]["tag"]) for x in buildOrder ]))
+           "\n - ".join([ x+" (development package)" if x in develPkgs else "%s@%s" % (x, specs[x]["tag"]) for x in buildOrder if x != "defaults-release" ]))
 
   if develPkgs:
     banner(format("You have packages in development mode.\n"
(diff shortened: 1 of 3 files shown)


ktf:
@dberzano @Barthelemy can you try this out?

Barthelemy:
I will, thank you for implementing this.

ktf:
You are welcome. Keep in mind that this will however trigger a rebuild of anything which had a default enabled.

ktf:
@sawenzel

dberzano:
@ktf I am testing it and... it started rebuilding everything from scratch (my installation had no default, meaning the `release` default). We should maybe have a look at it and understand why.

dberzano:
The problem is that if you have a list of environment variables defined in a default, they are no longer included in the (say) `init.sh` of the `defaults-whatever` package - they are rather added to each single package (the novel `.*` override).

Such variables are taken into account when computing hashes. So where previously there was no variable now you have some, and hashes change - therefore causing the rebuild.

dberzano:
@ktf can I test?

ktf:
Yes, please.

dberzano:
So, I have just checked two things:

1. no rebuild of current "release" software: check
2. global variables (which are now treated as overrides for the defaults-release package) are visible to each package's recipe: check

When building O2 with its defaults I have noticed that the "defaults-release" package still appears in the dependencies list (the "Packages will be built in the following order:" list).

Maybe for clarity I would filter it out there? I am opening a PR to this PR (!!!) to clarify.

ktf:
There is also still a problem with environment variables. FairRoot fails because ROOT does not define ROOTSYS.

ktf:
@dberzano @Barthelemy @sawenzel can you try this again?

ktf:
This is now merged in the dev branch.

PR #387, opened 2016-11-22
Add library relocation on OSX

dberzano:
Fix build problems and potentially allows to run with SIP enabled.

PR #420, opened 2017-02-13
Relocate libraries and binaries when running on Mac OSX.

fuhlig1:
When using aliBuild on Mac OSX the links of the library still point to the INSTALLROOT directory. Since this directory is normally removed after building the needed libraries can't be found at runtime. To overcome the problem add relocation command to the relocate-me.sh script which are executed after the package is installed to the final location.

### DIFF
--- a/alibuild_helpers/build_template.sh
+++ b/alibuild_helpers/build_template.sh
@@ -172,6 +172,45 @@ cat "$INSTALLROOT/relocate-me.sh"
 cat "$INSTALLROOT/.original-unrelocated" | xargs -n1 -I{} cp '{}' '{}'.unrelocated
 cd "$WORK_DIR/INSTALLROOT/$PKGHASH"
 
+if [[ $ARCHITECTURE =~ osx.* ]]; then
+  echo "   " >> "$INSTALLROOT/relocate-me.sh"
+  echo "# Add relocation of binaries and libraries when running on Mac OSX"  >> "$INSTALLROOT/relocate-me.sh"
+
+  if [ -d $INSTALLROOT/lib ]; then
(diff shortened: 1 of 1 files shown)


ktf:
Isn't #387 doing the same?

fuhlig1:
In my opinion there are two differences. 

#387 is only changing the name of the library, but not the dependencies to other libraries build in the same package. I think this is even mentioned in one of the comments.
The second problem is that the relocation is only done for libraries but not for the binaries. For example there is a problem if one tries to use the protoc tool without the
relocation. In the case without relocation the libraries are still searched in INSTALLROOT. 

Ciao

Florian

Am 13.02.2017 um 15:22 schrieb Giulio Eulisse <notifications@github.com>:

dennisklein:
I propose a small change: In order for the install names of the dylibs to actually have `@rpath` support, I believe we need to call `install_name_tool -id @rpath/libxxx.dylib` instead of the absolute path. I will test and verify it tomorrow.

@ktf @dberzano @fuhlig1 What else is needed to move this forward?

dberzano:
Hi all, could you have a closer look at #387 and make this PR look more like it, style-wise (but with the additional, missing functionalities)? In particular:

* There is an inconsistent use of `[[` and `[`
* There is a very strong assumption that all binaries are under `bin` and all libraries are under `lib`

dberzano:
Changes presented here have been ported almost as-is to #466, with a different logic to scan for directories containing relocatables, which is more lightweight (and it is configurable, but has a sensible default). Also, #466 does not rely on the (quite fragile) output of the `file` command but it scans directly the Mach-O headers using `otool -h`.

PR #444, opened 2017-05-03
prefetch git heads in parallel

dennisklein:
Some idea to speed up the start of the build phase. I have only tested it on python 2.7.13 (macOS 10.11), but it should work from 2.4+ on or so. It saves in the order of 15 seconds for me on larger dependency trees like `O2` or `FairRoot`.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -20,6 +20,7 @@ from alibuild_helpers.workarea import updateReferenceRepos
 from alibuild_helpers.log import logger_handler, LogFormatter, ProgressPrint, riemannStream
 from datetime import datetime
 from glob import glob
+from multiprocessing.pool import ThreadPool
 
 import socket
 import os
@@ -336,6 +337,33 @@ def doBuild(args, parser):
                   "  git pull --rebase\n",
                   pwd=os.getcwd(), star=star()))
 
+  # Prefetch git heads in parallel
+  debug("Prefetching git heads in parallel:")
+
+  fetchers = list(filter(lambda p: "source" in specs[p], buildOrder))
+  poolConfig = {"processes": min(max(len(fetchers), 1), 16)}
+  pool = ThreadPool(**poolConfig)
+  debug("Created thread pool with config: %s" % poolConfig)
+
+  git_commands = {} # store futures
+  for p in fetchers:
+    # Replace source with local one if we are in development mode.
+    if specs[p]["package"] in develPkgs:
+      specs[p]["source"] = join(os.getcwd(), specs[p]["package"])
+
+    cmd = "git ls-remote --heads %s" % specs[p]["source"]
+    debug("Enqueued '%s'" % cmd)
+    git_commands[p] = pool.apply_async(getstatusoutput, (cmd,))
+
+  results = {}
+  for p, future in git_commands.items():
+    results[p] = future.get()
+
+  for p, result in results.items():
+    dieOnError(result[0], "Unable to fetch from %s" % specs[p]["source"])
+    specs[p]["git_heads"] = result[1].split("\n")
+
+  debug("Finished prefetching git heads")
 
   # Resolve the tag to the actual commit ref, so that
   for p in buildOrder:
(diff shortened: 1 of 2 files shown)


alibuild_helpers/build.py
```diff
@@ -20,6 +20,7 @@
 from alibuild_helpers.log import logger_handler, LogFormatter, ProgressPrint, riemannStream
 from datetime import datetime
 from glob import glob
+from multiprocessing.pool import ThreadPool
```
ktf: Please add `multiprocessing` to the `setup.py` as well. While I agree this should always be there on any recent python distribution, it's better safe than sorry.

alibuild_helpers/build.py
```diff
@@ -336,31 +337,46 @@ def doBuild(args, parser):
...
+        specs[p]["source"] = join(os.getcwd(), specs[p]["package"])
+
+      cmd = "git ls-remote --heads %s" % specs[p]["source"]
+      debug("Enqueued '%s'" % cmd)
+      git_commands[p] = pool.apply_async(getstatusoutput, (cmd,))
+
+  for p in buildOrder:
+    if p in git_commands:
+      result = git_commands[p].get()
+      dieOnError(result[0], "Unable to fetch from %s" % specs[p]["source"])
```
ktf: Doesn't this leave potential zombies behind? Notice that we do not necessarily run in an environment where pid 1 is `init` so you cannot rely on the process reaper to be there. Can you change it so that you first do the for loop and then if there was an error you die?

dennisklein: `multiprocess.pool.ThreadPool` actually uses `multiprocess.dummy.Process` internally, which is not forking, but is implemented with the `threading` library.

alibuild_helpers/build.py
dennisklein: Done. But not sure, if it is doing, what you want, see https://pypi.python.org/pypi/multiprocessing/

alibuild_helpers/build.py
ktf: You still call getstatusoutput inside the thread, right? That will create the process which might end up being zombied.

alibuild_helpers/build.py
ktf: I agree, this is probably not what we want. Sorry for the noise.

dennisklein:
> What's the performance difference between 1, 4 and 32 fetchers?

Tested with `pb --defaults o2 build O2` which does **23 fetches in total** at the moment.

| # fetchers | wall time in seconds |
| --- | --- |
| 1 | 15.4 |
| 2 | 7.8 | 
| 4 | 4.1 |
| 8 | 2.1 |
| 16 | 1.3 |
| 32 | 0.9 |

ktf:
I'd make it 16, if not 8 (which is when your speedup starts to bend), the problem is that "limit" might be simply terminating the connection, which will kill the build.

alibuild_helpers/build.py
dennisklein: Reverted.

alibuild_helpers/build.py
dennisklein: True. I changed it. (Btw, I will cleanup and squash the git history after we are happy with the PR)

ktf: Thanks! Squashing at the end is fine.

dennisklein:
> I'd make it 16, if not 8 (which is when your speedup starts to bend)

Changed it to `16`. I suppose the bend here comes from the total number of fetches only being `23`. I would expect it to scale also to `32` and beyond with enough fetches, because the processes are just waiting for the network. But I am happy to go with `16`.

ktf:
Ok, great. Can you update the travis test to build a simple package which requires a couple of dependencies in parallel? Once that is done, it LGTM.

ktf:
LGTM, squash this and I will merge it.

dennisklein:
@ktf Can you pls review, if the test coverage is enough. I am not that familiar with the python unittest library. But I believe the new code paths are now at least executed.

ktf:
Race condition, I guess. Looks Good To Me, feel free to squash and I will merge it, alternatively update the PR description with the final commit message and I can do the squash myself.

dennisklein:
Hm, I have already squashed and force pushed to the same branch. I had to reload browser tab for github to show it though.

PR #445, opened 2017-05-04
improve osx relocation of dynamically linked binaries

dennisklein:
* search all directories
 * relocate `LC_ID_DYLIB`, `LC_RPATH` and `LC_LOAD_DYLIB` entries
 * relocate dylibs and executables

Co-authored-by: Dario Berzano <dario.berzano@cern.ch>
Co-authored-by: Florian Uhlig <f.uhlig@gsi.de>

---

This is part of the resolution of AliceO2Group/AliceO2#343. Moreover, this is based on the work done by @dberzano (#387) and @fuhlig1 (#420) - please read those discussions for context.

@dberzano 's change that invalidates existing builds, I have left out, because @ktf has reverted it in the `dev` branch, but I don't know the reason yet.

I did not implement my proposal about enforcing "`@rpath`iazation" that I have mentioned in the discussion in #420. After another discussion with @fuhlig1 he convinced me, that this should be an upstream decision.

The serial `find + file` variant takes several minutes on spinning disks and large packages like ROOT, that is why I chose a parallel variant.

The generated code could be more elegant, e.g. a variable with all the files and just one loop for each relocation type. But I found for debugging it is quite nice to have the code for single files in a copy&pastable format.

Please test! I only have one mac version `10.11.6` to test. You can inspect the resulting dylibs and executables with `otool -l`.

### DIFF
--- a/alibuild_helpers/build_template.sh
+++ b/alibuild_helpers/build_template.sh
@@ -168,14 +168,40 @@ cat "$INSTALLROOT/.original-unrelocated" | xargs -n1 -I{} echo "sed -e \"s|/[^ ]
 if [[ ! -s "$INSTALLROOT/.original-unrelocated" && -f "$INSTALLROOT/etc/modulefiles/$PKGNAME" ]]; then
   echo "mv -f \$PP/etc/modulefiles/$PKGNAME \$PP/etc/modulefiles/${PKGNAME}.forced-relocation && sed -e \"s|[@][@]PKGREVISION[@]\$PH[@][@]|$PKGREVISION|g\" \$PP/etc/modulefiles/${PKGNAME}.forced-relocation > \$PP/etc/modulefiles/$PKGNAME" >> "$INSTALLROOT/relocate-me.sh"
 fi
-# Find libraries needing relocation on macOS
-if [[ ${ARCHITECTURE:0:3} == osx ]]; then
-  find . -name '*.dylib' -o -name '*.so' | \
-    xargs -n1 -I{} sh -c "otool -D {}|tail -n1|grep -q $PKGHASH && echo {} | sed -e 's|^\.|\$PP|'" | \
-      while read LIB; do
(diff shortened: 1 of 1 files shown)


alibuild_helpers/build_template.sh
```diff
@@ -168,14 +168,40 @@ cat "$INSTALLROOT/.original-unrelocated" | xargs -n1 -I{} echo "sed -e \"s|/[^ ]
...
+cat << EOF >> "$INSTALLROOT/relocate-me.sh"
+install_name_tool -id \$(otool -D "\$PP/$BIN" | tail -n1 | sed -e "s|/[^ ]*INSTALLROOT/\$PH/\$OP|\$WORK_DIR/\$PP|g") "\$PP/$BIN"
+EOF
+    fi
+    # Relocate LC_RPATH
+    if { otool -l "$BIN" | grep -A2 LC_RPATH | grep path | grep -q $PKGHASH; }; then
+cat << EOF >> "$INSTALLROOT/relocate-me.sh"
+OLD_RPATHS=\$(otool -l \$PP/$BIN | grep -A2 LC_RPATH | grep path | grep \$PH | sed -e 's|^.*path ||' -e 's| .*$||')
+for OLD_RPATH in \$OLD_RPATHS; do
+  NEW_RPATH=\${OLD_RPATH/#*INSTALLROOT\/\$PH\/\$OP/\$WORK_DIR/\$PP}
```
dennisklein: In case you wonder, this is not a comment as suggested by github's syntax highlighting.., but a parameter substitution.

fuhlig1:
Hi  @dberzano, @ktf,
for me the PR looks okay, @dennisklein has nicely merged the two preexisting pull requests. Could you merge this one since more and more of our test users have problems on Mac OSX.

ktf:
For the record, we have discussed this today with @dberzano who has a solution which should be the sum of all the proposed ones. In particular we agreed on doing the fix based on paths matching a user specified regex (which defaults to `(lib|lib64)`). A little bit more patience and we will integrate the requested feature.

dennisklein:
Thank you very much, closing this one then.

PR #451, opened 2017-05-17
Remove special cases for setting mainPackage

### LINKED ISSUE #450: Get rid of the notion of mainPackage
It is the reason for misleading messages when building flpproto.

dberzano:
Fixes #450.


dberzano:
@ktf I believe it's not needed anymore, am I right? Any reason to keep these special cases?

ktf:
We probably still want the concept, but I agree the current implementation is flawed.

PR #457, opened 2017-07-11
Add documentation for O2 DAQ (flpproto)

dberzano:


dberzano:
@vascobarroso can you try out the documentation before merging? Thanks!

dberzano:
alisw/alidist#793 is related. In fact this doc should be used with alidist from that branch.

docs/_config.yml
```diff
@@ -22,6 +22,9 @@ menu:
   -
     name: o2-tutorial.html
     label: TUTORIAL (O2)
+  -
+    name: o2-daq-tutorial.html
+    label: TUTORIAL (O2 DAQ)
```
ktf: how about FLP PROTO rather than DAQ?

dberzano: It applies to all O2 DAQ packages. `flpproto`, as explained in the doc, is the "example" package (and by the way at some point it will be renamed to something else according to the authors).

Barthelemy:
+1

vascobarroso:
@dberzano documentation tried on a clean openstack node, works perfectly.
As mentioned via slack, the only change I would do at this point would be - given that they are development instructions - to suggest

```
aliBuild init flpproto@master
```

instead of 

```
aliBuild init flpproto
```

sy-c:
In the doc, it is written: 

Supported platforms
The whole O2 software supports the latest and long-term support versions of all
major Linux software distributions (RHEL/CentOS, Ubuntu, Fedora). In addition,
the two most recent releases of macOS are supported.

I don't think this is true for the hardware-related part... (rorc & readout)

dberzano:
Hi @sy-c, I am referring to a CWG11/WP3/O2 plenary discussion where this was, indeed, decided concerning the entirety of O2 software. Note that the policy (as WP3) is to _allow_ other groups to pick their favorite platform, among the ones we support. The reason why we support a broader range of platforms is that we wanted to make it easier for other groups to pick their favorite, so that the problem for us is no longer technical, but just a policy we leave up to the other groups.

You are probably referring to the fact that the software part of some hardware components will have, as per decision of other WPs, and signed off by WP3, **CC7 as their runtime platform**, which is the one we support the most. In the original document from our side, we say that we provide total support for CC7, and best-effort support for all the rest. If the hardware runs on CC7 nodes, there is really no need for us to even try compiling it on Ubuntu.

We did not elaborate in the tutorial, as it is, well, a tutorial, not a discussion on policies, and we do support multiple platforms for at least the vast majority of our recipes.

Maybe mitigating the sentence would help, let me do that :-)

Thanks for your feedback!

dberzano:
@sy-c I have added a note that explicitly mentions what is supported on all platforms, and what is not (notably, hardware drivers). This should prevent any confusion :-)

sy-c:
Thank you, this warning clarifies well our current HW support policy !

dberzano:
@Barthelemy @sy-c @vascobarroso 

You can now find the page at:

* http://alisw.github.io/alibuild/ (click on the left menu on **TUTORIAL (O2 DAQ)**)
* http://alisw.github.io/alibuild/o2-daq-tutorial.html (direct link)

PR #464, opened 2017-07-26
Add kernel devel RPMs to the O2 DAQ guide

dberzano:


dberzano:
@vascobarroso @Barthelemy we were still missing those packages too, for completeness (even if aliDoctor duly warns you)

PR #466, opened 2017-08-09
Add accumulated changes to alibuild

ktf:


dberzano:
For reference: the last thing we miss before merging this will be addressing the library relocation problem as discussed in #420, #387 and #445.

PR #467, opened 2017-08-21
Add check to prevent recursive alienv enter

dberzano:


dberzano:
@ktf see anything stupid?

PR #473, opened 2017-08-28
Add current package to log messages

dberzano:


dberzano:
@ktf I believe this is the only place where the log formatter needs to be changed. The result would be something like:

```
DEBUG:DDS:O2:...
```

when building the **DDS** package as part of an **O2** build.

PR #478, opened 2017-09-04
Fix various aliBuild messages

dberzano:


dberzano:
@ktf fixes only here :-)

dberzano:
@ktf thanks

PR #481, opened 2017-09-04
Add defaults validation to aliDoctor

dberzano:


dberzano:
This should address #480 for aliDoctor, I am not sure whether there are cleaner ways. The result is that user gets warned right away if the default is incorrect.

Note that `aliDeps` requires, on the other hand, deep refactoring (it does not honor defaults at all at the moment), but if it's not urgent... I would postpone it for the next release.

dberzano:
@ktf hold on I have found a better way to implement this.

dberzano:
Patch is ready for review and coverage's happy now.

PR #483, opened 2017-09-09
Use just bash, not /bin/bash

ktf:
Just found [a linux distro](https://nixos.org) where `/bin/bash` is not a valid assumption.


ktf:
Agreed. Let me do a better one.

ktf:
See #489

PR #490, opened 2017-10-09
Raise an error if a dependency has a different capitalization

ktf:


ktf:
Wait for merge. I think this can be done better...

ktf:
This way it should give a better error in case the package name is completely wrong.

davidrohr:
Hi, this fixes the problem I had, thanks.
Cheers

PR #495, opened 2017-10-31
Fix dylib relocation and support System Integrity Protection

syamasak:
Hi,
This commit fixes the leak of dylib relocations.

Previously fastjet build was failed with SIP due to the cgal dylibs are not correctly relocated.
Some dylibs which install_name only contain basename (no / in its name!) part are fixed to have abs path.
After this change, fastjet should build successfully.


ktf:
This is great, thanks.

ktf:
I gave a look at this and I was wondering if we cannot simply drop the path in the -id. Will that break SIP?

syamasak:
The `install_name` is copied to libraries which links it for locating the library, so we cannot empty it.
Because one of the SIP feature prevents DYLD_LIBRARY_PATH from working, os loader cannot find linked dylibs placed other than the standard path. e.g. /usr/lib

ktf:
Ok. Agreed this is a correct workaround. Given `$WORK_DIR` is already used elsewhere for the same purpose I guess this is just fine.

PR #496, opened 2017-11-07
Only clone reference repos by default, fetch updates with --fetch-repos or -u

### LINKED ISSUE #323: Feature request: offline mode
I would like to have the possibility to build everything as it is currently in my local copy without updating from remote repos. As the prime use case I see is the creation of a development copy from one of the local package clones. In that case it is useful to build exactly the same version again.

dennisklein:
This patch is quite useful when having only a slow or unreliable
Internet connection available, e.g. in trains or planes.

This patch changes the default behaviour of the `{build,init}` subcommands, so that they only clone the reference repos, if they are missing. The user has to pass `--fetch-repos` or `-u` to force fetching updates to the reference repos.

This patch reorganizes the `updateReferenceRepos` calls to before the build loop
and extracts the git heads from the local repos only. ~~If `--offline` flag
is passed the `updateReferenceRepos` calls are skipped.~~

---

Hope, we can take this as a start to resolve #323. What do you think? Still needs testing, just hacked this together half an hour ago ;) But it was really annoying to wait 20 seconds in the train for fetching the git heads on every build.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -22,7 +22,6 @@ from alibuild_helpers.workarea import updateReferenceRepos
 from alibuild_helpers.log import logger_handler, LogFormatter, ProgressPrint, riemannStream
 from datetime import datetime
 from glob import glob
-from multiprocessing.pool import ThreadPool
 try:
   from collections import OrderedDict
 except ImportError:
@@ -344,33 +343,22 @@ def doBuild(args, parser):
                   "  git pull --rebase\n",
                   pwd=os.getcwd(), star=star()))
 
-  # Prefetch git heads in parallel
-  debug("Prefetching git heads in parallel:")
+  # Clone/update repos
+  for p in [p for p in buildOrder if "source" in specs[p]]:
+    if not args.fetchRepos:
+      specs[p]["reference"] = join(abspath(args.referenceSources), p.lower())
+    if args.fetchRepos or not exists(specs[p]["reference"]):
+      updateReferenceRepos(args.referenceSources, p, specs[p])
 
-  fetchers = list(filter(lambda p: "source" in specs[p], buildOrder))
-  poolConfig = {"processes": min(max(len(fetchers), 1), 16)}
-  pool = ThreadPool(**poolConfig)
-  debug("Created thread pool with config: %s" % poolConfig)
-
-  git_commands = {} # store futures
-  for p in fetchers:
-    # Replace source with local one if we are in development mode.
+  # Retrieve git heads
+  for p in [p for p in buildOrder if "reference" in specs[p]]:
+    cmd = "git ls-remote --heads %s" % specs[p]["reference"]
     if specs[p]["package"] in develPkgs:
       specs[p]["source"] = join(os.getcwd(), specs[p]["package"])
-
-    cmd = "git ls-remote --heads %s" % specs[p]["source"]
-    debug("Enqueued '%s'" % cmd)
-    git_commands[p] = pool.apply_async(getstatusoutput, (cmd,))
-
-  results = {}
-  for p, future in git_commands.items():
-    results[p] = future.get()
-
-  for p, result in results.items():
-    dieOnError(result[0], "Unable to fetch from %s" % specs[p]["source"])
-    specs[p]["git_heads"] = result[1].split("\n")
-
-  debug("Finished prefetching git heads")
+      cmd = "git ls-remote --heads %s" % specs[p]["source"]
+    res, output = getStatusOutputBash(cmd)
+    dieOnError(res, "Error on '%s': %s" % (cmd, output))
+    specs[p]["git_heads"] = output.split("\n")
 
   # Resolve the tag to the actual commit ref, so that
   for p in buildOrder:
@@ -770,9 +758,6 @@ def doBuild(args, parser):
       debug(spec["cachedTarball"] and
             "Found tarball in %s" % spec["cachedTarball"] or
             "No cache tarballs found")
-
-    # FIXME: Why doing it here? This should really be done before anything else.
-    updateReferenceRepos(args.referenceSources, p, spec)
 
     # Generate the part which sources the environment for all the dependencies.
     # Notice that we guarantee that a dependency is always sourced before the
--- a/alibuild_helpers/init.py
+++ b/alibuild_helpers/init.py
@@ -5,7 +5,7 @@ from alibuild_helpers.log import debug, error, warning, banner, info
 from alibuild_helpers.log import dieOnError
 from alibuild_helpers.workarea import updateReferenceRepos
 
-from os.path import basename, join
+from os.path import abspath, basename, join
 import os.path as path
 import os
 try:
@@ -17,31 +17,31 @@ def parsePackagesDefinition(pkgname):
   return [ dict(zip(["name","ver"], y.split("@")[0:2]))
            for y in [ x+"@" for x in list(filter(lambda y: y, pkgname.split(","))) ] ]
 
-def doInit(setdir, configDir, pkgname, referenceSources, dist, defaults, dryRun):
-  assert(pkgname != None)
-  assert(type(dist) == dict)
-  assert(sorted(dist.keys()) == ["repo", "ver"])
-  pkgs = parsePackagesDefinition(pkgname)
+def doInit(args):
+  assert(args.pkgname != None)
+  assert(type(args.dist) == dict)
+  assert(sorted(args.dist.keys()) == ["repo", "ver"])
+  pkgs = parsePackagesDefinition(args.pkgname)
   assert(type(pkgs) == list)
-  if dryRun:
+  if args.dryRun:
     info("This will initialise local checkouts for %s\n"
          "--dry-run / -n specified. Doing nothing." % ",".join(x["name"] for x in pkgs))
     exit(0)
   try:
-    path.exists(setdir) or os.mkdir(setdir)
-    path.exists(referenceSources) or os.makedirs(referenceSources)
+    path.exists(args.develPrefix) or os.mkdir(args.develPrefix)
+    path.exists(args.referenceSources) or os.makedirs(args.referenceSources)
   except OSError as e:
     error(str(e))
     exit(1)
 
   # Fetch recipes first if necessary
-  if path.exists(configDir):
-    warning("using existing recipes from %s" % configDir)
+  if path.exists(args.configDir):
+    warning("using existing recipes from %s" % args.configDir)
   else:
     cmd = format("git clone %(repo)s%(branch)s %(cd)s",
-                 repo=dist["repo"] if ":" in dist["repo"] else "https://github.com/%s" % dist["repo"],
-                 branch=" -b "+dist["ver"] if dist["ver"] else "",
-                 cd=configDir)
+                 repo=args.dist["repo"] if ":" in args.dist["repo"] else "https://github.com/%s" % args.dist["repo"],
+                 branch=" -b "+args.dist["ver"] if args.dist["ver"] else "",
+                 cd=args.configDir)
     debug(cmd)
     err = execute(cmd)
     dieOnError(err!=0, "cannot clone recipes")
@@ -49,32 +49,32 @@ def doInit(setdir, configDir, pkgname, referenceSources, dist, defaults, dryRun)
   # Use standard functions supporting overrides and taps. Ignore all disables
   # and system packages as they are irrelevant in this context
   specs = {}
-  defaultsReader = lambda: readDefaults(configDir, defaults, error)
+  defaultsReader = lambda: readDefaults(args.configDir, args.defaults, error)
   (err, overrides, taps) = parseDefaults([], defaultsReader, debug)
   (_,_,_,validDefaults) = getPackageList(packages=[ p["name"] for p in pkgs ],
                                          specs=specs,
-                                         configDir=configDir,
+                                         configDir=args.configDir,
                                          preferSystem=False,
                                          noSystem=True,
                                          architecture="",
                                          disable=[],
-                                         defaults=defaults,
+                                         defaults=args.defaults,
                                          dieOnError=lambda *x, **y: None,
                                          performPreferCheck=lambda *x, **y: (1, ""),
                                          performRequirementCheck=lambda *x, **y: (0, ""),
-                                         performValidateDefaults=lambda spec : validateDefaults(spec, defaults),
+                                         performValidateDefaults=lambda spec : validateDefaults(spec, args.defaults),
                                          overrides=overrides,
                                          taps=taps,
                                          log=debug)
-  dieOnError(validDefaults and defaults not in validDefaults,
-             "Specified default `%s' is not compatible with the packages you want to build.\n" % defaults +
+  dieOnError(validDefaults and args.defaults not in validDefaults,
+             "Specified default `%s' is not compatible with the packages you want to build.\n" % args.defaults +
              "Valid defaults:\n\n- " +
              "\n- ".join(sorted(validDefaults)))
 
   for p in pkgs:
     spec = specs.get(p["name"])
     dieOnError(spec is None, "cannot find recipe for package %s" % p["name"])
-    dest = join(setdir, spec["package"])
+    dest = join(args.develPrefix, spec["package"])
     writeRepo = spec.get("write_repo", spec.get("source"))
     dieOnError(not writeRepo, "package %s has no source field and cannot be developed" % spec["package"])
     if path.exists(dest):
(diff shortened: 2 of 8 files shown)


ktf:
In principle we do not really need to fetch that information and we can simply assume what is there is ok. Maybe `--offline` should simply extract the same information from the `.git` repository, rather than storing it in a separate file?

dennisklein:
> Maybe --offline should simply extract the same information from the .git repository, rather than storing it in a separate file?

I moved the repo clone/updates in front of the `git ls-remote` calls as discussed last week. This way we can do the `git ls-remote --heads` calls always against the local repos. What do you think?

In principle one could apply the parallel fetching now to the repo clone/update loop. But it seems people have problems with it (#498), so I left it out for now.

alibuild_helpers/build.py
```diff
@@ -344,33 +344,30 @@ def doBuild(args, parser):
...
-    # Replace source with local one if we are in development mode.
+  # Clone/update repos
+  reposWithSource = list(filter(lambda p: "source" in specs[p], buildOrder))
+  if args.offline:
+    banner("Offline mode - not attempting to update any repositories\n")
+    for p in reposWithSource:
+      specs[p]["reference"] = "%s/%s" % (abspath(args.referenceSources), p.lower())
+      dieOnError(not exists(specs[p]["reference"]), "Local clone of %s does not exist yet, run again without --offline option.");
+  else:
+    banner("Cloning/Updating repositories\nUse --offline to skip this step. Skipping will only work after the first\nsuccessful completion of this step without --offline.\n")
```
ktf: Do we really need this printed every time as a banner?

alibuild_helpers/build.py
```diff
@@ -344,33 +344,30 @@ def doBuild(args, parser):
...
-    results[p] = future.get()
-
-  for p, result in results.items():
-    dieOnError(result[0], "Unable to fetch from %s" % specs[p]["source"])
-    specs[p]["git_heads"] = result[1].split("\n")
-
-  debug("Finished prefetching git heads")
+      cmd = "git ls-remote --heads %s" % specs[p]["source"]
+    res, output = getstatusoutput(cmd)
+    if sys.version_info < (3, 0): # pragma: no cover
```
ktf: Can you be a more detailed about what "pragma: no cover" means?

tests/test_build.py
```diff
@@ -88,8 +88,8 @@ def dummy_getstatusoutput(x):
       'pip --disable-pip-version-check show alibuild | grep -e "^Version:" | sed -e \'s/.* //\'': (0, "v1.5.0"),
       'which pigz': (1, ""),
       'tar --ignore-failed-read -cvvf /dev/null /dev/zero': (0, ""),
-      'git ls-remote --heads https://github.com/root-mirror/root': (0, "87b87c4322d2a3fad315c919cb2e2dd73f2154dc\trefs/heads/master\nf7b336611753f1f4aaa94222b0d620748ae230c0\trefs/heads/v6-08-00-patches"),
-      'git ls-remote --heads https://github.com/star-externals/zlib': (0, "8822efa61f2a385e0bc83ca5819d608111b2168a\trefs/heads/master")
+      'git ls-remote --heads /sw/MIRROR/root': (0, "87b87c4322d2a3fad315c919cb2e2dd73f2154dc\trefs/heads/master\nf7b336611753f1f4aaa94222b0d620748ae230c0\trefs/heads/v6-08-00-patches"),
```
ktf: What will happen of the non `--offline` tests?

alibuild_helpers/build.py
dennisklein: My thinking of making it a banner was, that it is a step that takes a long time. At least in the inverted case (offline mode enabled), @sawenzel suggested to make it very clear in the output.

I leave the decision to you, I don't mind printing in a different severity level.

alibuild_helpers/build.py
dennisklein: It omits this branch from the coverage report. It is only needed, if the coverage reports are generated with python 3+, because in that case the branch will never be executed. I have not verified, if it works though. See http://coverage.readthedocs.io/en/coverage-4.3.4/excluding.html

alibuild_helpers/build.py
ktf: Please leave it out for now. We should leave the old behavior unchanged for the moment.

alibuild_helpers/build.py
ktf: We do run tests on both python 2.7 and 3.6, so in principle we should be able to get coverage for both. Can you please leave it out?

tests/test_build.py
dennisklein: In both offline and online mode, it will lookup the git heads locally.

Before the logic was:
1. Lookup git heads remotely (slow)
2. Decide, if we need to build
3. As part of the build update local repo (slow)
Cost: N git heads lookups + M repo updates, if build is needed
Best case: N (no build needed), worst case: 2N

Now it is:
1. Update local repos (slow)
2. Lookup git heads locally (fast)
3. Decide, if we need to build
Cost: N repo updates

Repo updates and remote git head lookups cost about the same in terms of wall time.

This PR is a regression in total wall time for online mode though, because there is no parallelization any more on the slow network requests that are executed on every build invocation. But as I said, I could easily add the parallelization, but there is an open issue about this.

alibuild_helpers/build.py
dennisklein: Done

alibuild_helpers/build.py
dennisklein: Done

dennisklein:
Also updated the docs. Hope, it is what you had in mind.

ktf:
Perfect, thanks.

ktf:
@matthiasrichter can you check if this addresses your feature request?

matthiasrichter:
Hi, yes, this is pretty much what I had in mind, thanks @dennisklein for embarking on this. I had a similar experience as you on a slow internet connection.

dennisklein:
@matthiasrichter While reading your issue, I noticed, you had also an additional use case in mind. You wanted to initialize a devel package from local clones. I have added this just now in the second commit.

alibuild_helpers/build.py
```diff
@@ -344,33 +344,29 @@ def doBuild(args, parser):
...
-    results[p] = future.get()
-
-  for p, result in results.items():
-    dieOnError(result[0], "Unable to fetch from %s" % specs[p]["source"])
-    specs[p]["git_heads"] = result[1].split("\n")
-
-  debug("Finished prefetching git heads")
+      cmd = "git ls-remote --heads %s" % specs[p]["source"]
+    res, output = getstatusoutput(cmd)
+    if sys.version_info < (3, 0):
```
dberzano: We have a `to_unicode` function in utilities, and a `getStatusOutputBash` that should be used instead (no need to check for Python versions here nor to `decode`).

alibuild_helpers/build.py
```diff
@@ -344,33 +344,29 @@ def doBuild(args, parser):
...
-  debug("Prefetching git heads in parallel:")
-
-  fetchers = list(filter(lambda p: "source" in specs[p], buildOrder))
-  poolConfig = {"processes": min(max(len(fetchers), 1), 16)}
-  pool = ThreadPool(**poolConfig)
-  debug("Created thread pool with config: %s" % poolConfig)
+  # Clone/update repos
+  reposWithSource = list(filter(lambda p: "source" in specs[p], buildOrder))
+  if args.offline:
+    banner("Offline mode - not attempting to update any repositories\n")
```
dberzano: I would not be so verbose here and change the style of the output:

```
debug("Not fetching repository updates")
```

will do :-)

alibuild_helpers/init.py
```diff
@@ -82,7 +82,14 @@ def doInit(setdir, configDir, pkgname, referenceSources, dist, defaults, dryRun)
       continue
     p["ver"] = p["ver"] if p["ver"] else spec.get("tag", spec["version"])
     debug("cloning %s%s for development" % (spec["package"], " version "+p["ver"] if p["ver"] else ""))
-    updateReferenceRepos(referenceSources, spec["package"], spec)
+
+    if offline:
+      debug("Offline mode - not attempting to update any repositories");
```
dberzano: Same comment about message and please drop the semicolon :-)

alibuild_helpers/init.py
```diff
@@ -82,7 +82,14 @@ def doInit(setdir, configDir, pkgname, referenceSources, dist, defaults, dryRun)
...
+      spec["reference"] = "%s/%s" % (abspath(referenceSources), spec["package"].lower())
+      dieOnError(not path.exists(spec["reference"]), "Local clone of %s does not exist yet, run again without --offline option." % spec["package"]);
```
dberzano: Drop `;`

docs/user.markdown
```diff
@@ -62,6 +63,8 @@ For a quick start introduction, please look [here](./quick.html).
                             recipes set ([user/repo@]branch)
       --dry-run, -n         Prints what would happen, without actually doing the
                             build.
+      --offline             Offline mode. Does not attempt to connect to the
+                            Internet to fetch repositories.
```
dberzano: No `.` at the end of the sentence (incidentally, it has to be changed on `--dry-run` too)

docs/user.markdown
```diff
@@ -248,3 +251,19 @@ or in case you need to be root (_e.g._ on Ubuntu and most Linux distributions
 for convenience):
 
     sudo pip install --upgrade alibuild
+
+## Offline mode
+
+If one needs to frequently invoke the `aliBuild build` subcommand, it can be
```
dberzano: The whole paragraph can be more concise IMHO.

dberzano:
@mpuccio is interested in this PR too.

dennisklein:
@dberzano Thx for your feedback. I will try to get it implemented with the reversed flag logic as you described.

Another thing I noticed is that the current implementation triggered a (almost?) full rebuild on my setup. Maybe one of the spec fields I modify is sensible to the hash calculation? On the other hand, the unit tests do not complain..

dennisklein:
Updated to the latest idea. Now getting different build hashes from your CI too. At least, they are consistent with the ones on my machine. Will have to look into this tomorrow..

ktf:
How come you get different hashes? That should definitely not happen.


Quoting Dennis Klein (2017-11-21 20:34:39)

dennisklein:
> How come you get different hashes? That should definitely not happen.

Hehe, I switched to `getStatusOutputBash` as suggested by @dberzano and it was not yet mocked in the tests..

@dberzano @ktf Once you are happy with the implementation and you give it a go, I will fix the test coverage.

dennisklein:
oh, I am confused by the test coverage report. When I push to the PR branch, it first reports terrible test coverage and after a while it updates again and shows better results..

ktf:
What's up with this?

alibuild_helpers/build.py
```diff
@@ -344,33 +344,30 @@ def doBuild(args, parser):
...
-  poolConfig = {"processes": min(max(len(fetchers), 1), 16)}
-  pool = ThreadPool(**poolConfig)
-  debug("Created thread pool with config: %s" % poolConfig)
-
-  git_commands = {} # store futures
-  for p in fetchers:
-    # Replace source with local one if we are in development mode.
+  # Clone/update repos
+  reposWithSource = list(filter(lambda p: "source" in specs[p], buildOrder))
+  if args.fetchRepos:
```
dberzano: This whole `if...else` can be simplified I believe (also, use `join` not `%s/%s`):

```python
for p in list(filter(lambda p: "source" in specs[p], buildOrder)):
  if not args.fetchRepos:
    specs[p]["reference"] = os.path.join(args.referenceSources, p.lower())
  if args.fetchRepos or not exists(spec[p]["reference"]):
    updateReferenceRepos(args.referenceSources, p, specs[p])
```

alibuild_helpers/build.py
```diff
@@ -344,33 +344,30 @@ def doBuild(args, parser):
...
+      updateReferenceRepos(args.referenceSources, p, specs[p])
+  else:
+    # Only create reference if it does not exist
+    for p in reposWithSource:
+      specs[p]["reference"] = "%s/%s" % (abspath(args.referenceSources), p.lower())
+      if not exists(specs[p]["reference"]):
+        updateReferenceRepos(args.referenceSources, p, specs[p])
+
+  # Retrieve git heads
+  git_head_fetchers = list(filter(lambda p: "reference" in specs[p], buildOrder))
```
dberzano: Here too we can have all in one line:

```python
for p in list(filter(lambda p: "reference" in specs[p], buildOrder)):
  ...
```

alibuild_helpers/build.py
```diff
@@ -344,33 +344,30 @@ def doBuild(args, parser):
...
-    results[p] = future.get()
-
-  for p, result in results.items():
-    dieOnError(result[0], "Unable to fetch from %s" % specs[p]["source"])
-    specs[p]["git_heads"] = result[1].split("\n")
-
-  debug("Finished prefetching git heads")
+      cmd = "git ls-remote --heads %s" % specs[p]["source"]
+    res, output = getStatusOutputBash(cmd)
+    output = to_unicode(output)
```
dberzano: I believe output of `git ls-remote` is ASCII, I think we don't need to convert it to Unicode. Besides, `getStatusOutputBash()` already returns Unicode, so this conversion can be dropped.

alibuild_helpers/init.py
```diff
@@ -17,7 +17,15 @@ def parsePackagesDefinition(pkgname):
   return [ dict(zip(["name","ver"], y.split("@")[0:2]))
            for y in [ x+"@" for x in list(filter(lambda y: y, pkgname.split(","))) ] ]
 
-def doInit(setdir, configDir, pkgname, referenceSources, dist, defaults, dryRun):
```
dberzano: I'd leave it as it was, we're redefining variables for no reason.

alibuild_helpers/init.py
```diff
@@ -82,7 +90,14 @@ def doInit(setdir, configDir, pkgname, referenceSources, dist, defaults, dryRun)
       continue
     p["ver"] = p["ver"] if p["ver"] else spec.get("tag", spec["version"])
     debug("cloning %s%s for development" % (spec["package"], " version "+p["ver"] if p["ver"] else ""))
-    updateReferenceRepos(referenceSources, spec["package"], spec)
+
+    if fetchRepos:
```
dberzano: Here again I would simplify it. The action `updateReferenceRepos` IMHO should be semantically expressed only once, with a careful specification of the conditions that lead to it. See my example for `build.py`.

dberzano:
I would say that when we address the remaining little issues we can merge, and make a `rc`.

alibuild_helpers/build.py
brinick: Would it not be more idiomatic to use a list comprehension?
e.g.
```for p in [p for p in buildOrder if "reference" in specs[p]]:```

Comment (dberzano) [reply alibuild_helpers/build.py]: Even better, yes, I fully agree.

File: alibuild_helpers/build.py (hunk 10 above)
Comment (brinick) [reply alibuild_helpers/build.py]: List comprehensions for the win!
```python
for p in [p for p in buildOrder if "source" in specs[p]]:
```

dberzano: 👍!

alibuild_helpers/build.py
dennisklein: I had problems with the `Vc` repo here. Seems the output can be non ASCII. But even better, if `getStatusOutputBash()` takes care of this :)

dberzano: Damn, I have tried it and you are right!

```
git ls-remote --heads https://github.com/VcDevel/Vc.git | iconv -futf-8 -tascii
```

ends with:

```
667f4f7bd1facd6783270ebf38f2a54a7e6bc524	refs/heads/mkretz/gh-102_AVX512
9a0d1481502f81bab311be6afa0d086e631f4df3	refs/heads/mkretz/gh-14iconv: illegal input sequence at position 994
```

What the hell man!

alibuild_helpers/init.py
dennisklein: Do you insist on reverting this change? I had to add a variable to the argument list and thought it is getting too long. So I followed the example of `doBuild`, because all arguments are a member of the `args` dictionary already. This has also an impact on the calls in the unit test files.

dberzano: Say, in any case it does not pertain to the topic of this PR (neither does adding Fedora to the list of supported platforms!) I would revert it for now (we need to get the nice feature you've developed quickly available), and if I implemented it I would not duplicate variables like this (there's plenty of `blah = args.blah`, why not accessing `args.blah` directly then?)

dberzano: (Just leave the Fedora part there, it's fine...)

dennisklein: > why not accessing `args.blah` directly then?

Done :P

dberzano:
Thanks @dennisklein for the hard work! I am having a last quick look before proceeding.

dberzano:
I have run the following two tests:

- [X] Try to see if `--dry-run` is honored
- [X] Force-delete a tag from MIRRORS and SOURCES and see if it recovers

I'd say we can finally merge this, this is a very useful feature!

ktf:
This is breaking daily builds:

https://alijenkins.cern.ch/job/DailyBuilds/job/DailyO2/17/console

I suspect the problem is that it does not handle read-only repos. Unless you quickly find a solution, I will have to revert this.

dberzano:
@ktf is right. Note that CI jobs are (temporarily) running with a pinpointed alibuild version so there is no need to revert, but it needs to be fixed quickly...

ktf:
Can we pinpoint O2 builds as well?

PR #500, opened 2017-11-15
Fix ignored --debug in aliDoctor (and minor typos)

dberzano:


ktf:
We do use `arg.debug`, is it not honoured?

ktf:
E.g. at

https://github.com/dberzano/alibuild/blob/d71cf1f120e3913aeedd617cd5892ccb4cc91316/alibuild_helpers/doctor.py#L141

dberzano:
For some reason it is always `False` in case there is a parser, and a subparser, with the same argument name. I suspect this is not supported, and the `--debug` switches the value of the parser's argument, but `args` eventually contains the value of the subparser's, which is still set to `False`.

The message is that we should not have the same switch defined in both parser and subparsers.

PR #501, opened 2017-11-16
Force use of /bin/bash when checking system deps

dberzano:
(Also move Bash detection to cmd module and cache result.)

This fixes a serious problem when checking system packages: we assume the syntax is Bash's (therefore we are not POSIX-compliant), but we cannot rely on Python's shell choice. By default it's `/bin/sh`, which on _some_ Ubuntu systems is `/bin/bash` (OK), but on a large number it's `/bin/dash` (POSIX: not OK for us).

We need to make sure checks are executed in Bash.

Code for automatic detection of Bash (`/bin/bash`, `bash`) has been reused.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -10,7 +10,7 @@ except ImportError:
 from alibuild_helpers.analytics import report_event
 from alibuild_helpers.log import debug, error, info, banner, warning
 from alibuild_helpers.log import dieOnError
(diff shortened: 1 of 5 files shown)


dberzano:
Probably, coverage will complain. Before adding proper tests I'd like to converge on this. The problem seems serious and in some cases it leads to `aliDoctor` detecting a system package which is actually not there.

dberzano:
@lbariogl this is a potential solution to your problem.

PR #503, opened 2017-11-28
Make build passing with SIP

syamasak:
Still does not work with alienv yet.

### DIFF
--- a/alibuild_helpers/build_template.sh
+++ b/alibuild_helpers/build_template.sh
@@ -104,13 +104,13 @@ cd "$BUILDDIR"
 #   the relocation script so that it takes into account the new location.
 if [[ "$CACHED_TARBALL" == "" && ! -f $BUILDROOT/log ]]; then
   set -o pipefail
-  bash -ex "$WORK_DIR/SPECS/$ARCHITECTURE/$PKGNAME/$PKGVERSION-$PKGREVISION/$PKGNAME.sh" 2>&1 | tee "$BUILDROOT/log"
+  (set -x; echo KKKKK $LD_LIBRARY_PATH; source "$WORK_DIR/SPECS/$ARCHITECTURE/$PKGNAME/$PKGVERSION-$PKGREVISION/$PKGNAME.sh" 2>&1) | tee "$BUILDROOT/log"
 elif [[ "$CACHED_TARBALL" == "" && $INCREMENTAL_BUILD_HASH != "0" && -f "$BUILDDIR/.build_succeeded" ]]; then
   set -o pipefail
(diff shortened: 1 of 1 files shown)


alibuild_helpers/build_template.sh
```diff
@@ -104,13 +104,13 @@ cd "$BUILDDIR"
 #   the relocation script so that it takes into account the new location.
 if [[ "$CACHED_TARBALL" == "" && ! -f $BUILDROOT/log ]]; then
   set -o pipefail
-  bash -ex "$WORK_DIR/SPECS/$ARCHITECTURE/$PKGNAME/$PKGVERSION-$PKGREVISION/$PKGNAME.sh" 2>&1 | tee "$BUILDROOT/log"
+  . "$WORK_DIR/SPECS/$ARCHITECTURE/$PKGNAME/$PKGVERSION-$PKGREVISION/$PKGNAME.sh" 2>&1 | tee "$BUILDROOT/log"
```
dberzano: Hi @syamasak, we deliberately open a new Bash shell as we want a new environment. Sourcing the build script inside the current one is not exactly an option. We absolutely don't want a recipe to taint the rest of the environment. On top of that, where's `set -x` here?

However, I see an easy solution to that. You can try to amend this PR by wrapping script calls in `(...)`, so that you are effectively opening a new subshell with its separate environment.

Something like (and please use `source`, it's clearer than `.`):

```bash
 (set -x; source "$WORK_DIR/SPECS/$ARCHITECTURE/$PKGNAME/$PKGVERSION-$PKGREVISION/$PKGNAME.sh" 2>&1) | tee "$BUILDROOT/log"
```

The subshell should exit with an errorcode, and since we have already `set -o pipefail` and `set -e` this should cause the build template to exit with an error too.

Can you please implement my suggestion, and check if it effectively works by deliberately breaking a build recipe please?

Thanks!

syamasak:
Hi @dberzano, thanks for your great suggestions!

I haven't noticed that `bash -ex` works as a sandbox with trace, and a bit surprised `bash` does not spawn a subprocess for the simple case. After some tries, I managed to build AliPhysics.

I confirmed the recipe does not pollute the parent environment and `exit 1` in the recipe cause the build template to exit as expected.

AliRoot is now working with `eval "$(alienv load AliRoot/latest)"` :)

ktf:
What's the status of this? If I understand correctly, things works, but do we need anything to be merged upstream?

syamasak:
I've rebased this commit to the master and AliPhysics worked locally.
This commit is independent so it should work without any upstream change.
By the way, I don't know how to improve codecov/project status. Could you give me an idea?

PR #507, opened 2017-12-08
Allow read-only references

brinick:
If a mirror is read-only, ignore it and move on, otherwise fetch
the heads.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -372,7 +372,7 @@ def doBuild(args, parser):
 
     specs[p]["git_heads"] = output
 
-  # Resolve the tag to the actual commit ref, so that
+  # Resolve the tag to the actual commit ref
   for p in buildOrder:
     spec = specs[p]
     spec["commit_hash"] = "0"
@@ -980,7 +980,7 @@ def doBuild(args, parser):
 
     buildErrMsg = format("Error while executing %(sd)s/build.sh on `%(h)s'.\n"
                          "Log can be found in %(w)s/BUILD/%(p)s-latest%(devSuffix)s/log.\n"
-                         "Please upload it to CernBox/DropBox if you intend to request support.\n"
+                         "Please upload it to CERNBox/Dropbox if you intend to request support.\n"
                          "Build directory is %(w)s/BUILD/%(p)s-latest%(devSuffix)s/%(p)s.",
                          h=socket.gethostname(),
                          sd=scriptDir,
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -18,7 +18,7 @@ from alibuild_helpers.utilities import validateDefaults
 from alibuild_helpers.utilities import Hasher
 from alibuild_helpers.utilities import yamlDump
 import yaml
-from alibuild_helpers.workarea import updateReferenceRepos, is_writeable
+from alibuild_helpers.workarea import updateReferenceRepos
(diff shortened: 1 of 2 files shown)


alibuild_helpers/build.py
```diff
@@ -352,13 +351,26 @@ def doBuild(args, parser):
...
     if specs[p]["package"] in develPkgs:
       specs[p]["source"] = join(os.getcwd(), specs[p]["package"])
       cmd = "git ls-remote --heads %s" % specs[p]["source"]
-    res, output = getStatusOutputBash(cmd)
-    dieOnError(res, "Error on '%s': %s" % (cmd, output))
-    specs[p]["git_heads"] = output.split("\n")
+
+    output = []
+    if cmd:
+      res, output = getStatusOutputBash(cmd)
```
dberzano: Hmm I would simplify it much more by changing the command, that'd be a one-line modification:

```
cmd = format("touch %(ref)s && git ls-remote --heads %(ref)s || true", ref=specs[p]["source"])
```

This approach has two advantages:

* It actually _tries_ to write instead of trusting the `w` bit (which might not be representative of the actual state, see read-only filesystems, or filesystems which are read-only but can _appear_ to be `rw` like CVMFS or many FUSE modules).
* That'd be a one-line modification.

Concerning actually trying to write, even if it's implemented in Bash, it's more "idiomatic": we try to do something and then we say sorry if we couldn't, instead of checking if we can.

What do you think?

dberzano:
I know I am picky, sorry!

alibuild_helpers/build.py
brinick: yeah, ok :wink:
i will push a commit

brinick: Actually, do we need the ```|| true```?
Without it, if either the touch or git commands fail,
getBashStatusOutput simply returns a non-zero exit code
along with the error message.

dennisklein:
Why can we not do `git ls-remote --heads ...`, if the reference repo is not writable?

How is this change related to the failing daily builds? `updateReferenceRepos` is not performing any updates due to missing write access and does not find any existing reference repos in the given directory. What am I missing?

dennisklein:
If you have a look in the other daily build logs `updateReferenceRepo` fails with the same error. The difference is, before my patch, `git ls-remote --heads ...` was done against the github repos. Now, these commands run always against the local clone. Omitting is not an option. I believe we should put some actual bare repos into `/build/mirror` on the build hosts or fall back to looking up the git heads online if no local reference clone is found.

ktf:
The local clone is not guaranteed to be there, so you need to be able to fallback. `/build/mirror` might or might not exists depending on the deployments and there is no guarantees about it.

dennisklein:
What do you think of this solution:

```diff
diff --git a/alibuild_helpers/workarea.py b/alibuild_helpers/workarea.py
index 3945bd6..bbea46e 100644
--- a/alibuild_helpers/workarea.py
+++ b/alibuild_helpers/workarea.py
@@ -38,6 +38,7 @@ def updateReferenceRepos(referenceSources, p, spec):
     return
   if not writeableReference:
     debug("Cannot create reference for %s in specified folder.", p)
+    spec["reference"] = spec["source"]
     return
 
   err, out = getstatusoutput("mkdir -p %s" % abspath(referenceSources))
```

dberzano:
Note that we've merged it by accident, we'll be taking care of the failing tests before reopening (but the functionality is finally there).

PR #511, opened 2018-01-16
Fix tests

dberzano:


dberzano:
Let's continue what we've left behind in #507.

dberzano:
Just to be sure we are all on the same page: this PR should finalize all the problems we've encountered with #496 + #507, and improve it.

In case a recipe has a `tag:` pointing to a branch (very unlikely; it's however the case for FairRoot in some defaults, for instance), aliBuild needs to perform a remote operation (`git fetch`) in order to determine what's the branch's hash.

In case the package was previously built already, such piece of information is used to determine whether it should be rebuilt or not.

Unfortunately, many `git fetch` operations slow down aliBuild's startup, and are a huge annoyance to users performing small changes and incremental rebuilds.

@dennisklein's original idea (#444) was to fetch in parallel, but in some cases (read: github.com) this might fail on the server's side, probably because too many requests in parallel are spotted as DDoSes.

Current implementation sacrifices consistency in favour of speed.

* Fetch occurs only the first time you run aliBuild. This leaves a cached bare clone, by default under `sw/MIRROR/package`.
* Subsequent calls will never update that bare clone. Correspondence between hashes and branches will be read from there, and can be outdated. Since our production recipes use tags, not branches, this is not an issue on user laptops, and improves the overall speed.
* In case update is really necessary, it can be forced with the `-u|--fetch-repos` option.

All in all this change should address some of our user's problems: @mpuccio and @lbariogl can finally rebuild quickly, without random "fetch failures". @matthiasrichter should be able to rebuild while travelling without Internet connection :-)

docs/user.markdown
```diff
@@ -258,3 +258,19 @@ update of `alidist`, which users will have to be done separately.
...
+## Fetching updates
+
+aliBuild tries to determine whether a package should be rebuilt or not. In case
+the package's version (`tag:`) is a _branch_ name, aliBuild needs to query the
+remote repository to see if the remote branch has changed.
+
+Since the vast majority of recipes do not use branches as versions (they use
+actual, fixed tags instead), aliBuild does not perform any remote repository
+query by default, and uses previously cached information. This results in
+reduced network activity and an improved startup time.
```
ktf: Ok, but what happens if I add a new tag? Will that be transparently fetched?

dberzano: Yes because this logic is inside `build_template.sh`.

dberzano:
@ktf what about having the `--fetch-repos` doc in `troubleshooting` instead of `user`? Our `reference` doc is recipe-centric, does not really explain what the options do.

PR #516, opened 2018-01-29
Ensure we fetch ALL heads when updating reference

dberzano:
`git fetch` does not fetch all upstream branches by default, just the HEAD


dberzano:
@ktf this is very critical (it affected, negatively, AliDPG builds) and I think we should merge & patch-tag (after proper verification from you). This is the only way to tell `git fetch` to fetch all heads as intended, otherwise it will just fetch the `HEAD` (default branch). Without this patch, `aliBuild --fetch-repos` does not behave as intended.

ktf:
Can you add AliDPG builds to our tests?

PR #521, opened 2018-02-19
Rename o2-daq to o2-dataflow

dberzano:


dberzano:
@ktf this commit also fixes the continuous-yet-random problems we've been having with HTMLProofer, please have a look!

ktf:
@graeme-a-stewart you were telling me about similar issues with `HTMLProofer` a while ago, how did you end up solving them?

graeme-a-stewart:
We do have some issues with timeouts, but that's more because of annoyingly flaky websites we have some links too. For the really persistently bad ones we did even add the `{:data-proofer-ignore=""}` attribute and just ignore them (kind of defeats the purpose, but it solved some always red failures). It's rather annoying that the `htmlproofer` command line doesn't expose the timeout options that @dberzano used.

The other issue we had was sites where the intermediate certificates were missing. This we solved my adding these certificates into our own docker container as it was built:

https://github.com/HEP-SF/docker/blob/master/hsf-jeykll/Dockerfile#L4

dberzano:
@graeme-a-stewart yes, `htmlproofer` does not expose cmdline options by choice. They even say that the _only_ way to customize such options is to use the Ruby class directly as I did.

Once fixed the timeout issues, we never had CA problems, but we had more general SSL problems instead. Apparently, HTMLProofer has troubles using the GnuTLS libraries (installed by default on the Travis environment), and the solution is simply to switch to OpenSSL (hence the `apt install libcurl4-openssl-dev` in the `.travis.yml`).

PR #522, opened 2018-02-28
Do not allow custom grep options

ktf:


ktf:
@awegrzyn can you confirm this would have solved part of your issues with building ROOT?

awegrzyn:
@ktf Yes, it solves my issue.

ktf:
Errors unrelated.

PR #523, opened 2018-03-21
Disable shell hash command as hashing is disabled

dberzano:
When using `set +h`, the `hash` command returns nonzero with a warning saying
that hashing is disabled. To prevent recipes using `hash` to fail without
changing them, the `hash` command is being aliased to a no-op.

Note that this was introduced in #503 since:

* hashing is disabled in the `build_template.sh` shell environment,
* the recipe turned from being executed in a separate shell to being sourced in
  the current shell (thus retaining the `set +h` option).

### DIFF
--- a/alibuild_helpers/build_template.sh
+++ b/alibuild_helpers/build_template.sh
@@ -4,6 +4,7 @@
 
 set -e
 set +h
+function hash() { true; }
 export WORK_DIR="${WORK_DIR_OVERRIDE:-%(workDir)s}"
 
 # From our dependencies


dberzano:
@ktf yes, we've introduced in 2015(!) because some recipes failed using just-generated binaries (as some old binary was previously hashed by the shell, and therefore cached). Turns out, until #503 our `set +h` option was in any case never effective (except for incremental recipes), and now it is, but I have found out that, unfortunately, when this option is in place `hash -r` fails with:

```
$> hash -r
-bash: hash: hashing disabled
```

returning nonzero. We have some `hash -r` in the GCC recipe... luckily this is not in a tagged version.

PR #533, opened 2018-04-12
Do not prune _VERSION vars if not in alienv

### LINKED ISSUE #530: alibuild_helpers.utilities.prunePaths undefines essential Cray PE environment variables
Running `aliBuild build` on Cray XC40, with Cray PE installed, fails with missing packages that are however properly installed. And interestingly, `aliBuild doctor` is able to successfully pick up those packages. Porting the diagnostic code from `alibuild_helpers.doctor` into `alibuild_helpers.build` reveals that the failure is (and similarly, for `Xdevel`, `curl`, and `libperl`):

```
ERROR: Package opengl is a system requirement and cannot be found.
ERROR: This is due to the fact that the following script fails:
ERROR: 
ERROR: brew() { true; }; printf "#include <GL/glu.h>\n" | cc -xc - -c -o /dev/null
ERROR: with the following output:
ERROR: 
ERROR: opengl: Error:
ERROR: opengl: Unable to determine compiler version.
ERROR: opengl: Make sure that a gnu module is loaded and that GNU_VERSION is defined
ERROR: opengl: 
ERROR: OpenGL development packages are missing on your system.
ERROR:   * On RHEL-compatible systems you probably need: mesa-libGLU-devel
ERROR:   * On Ubuntu-compatible systems you probably need: libglu1-mesa-dev
ERROR: 
ERROR: 
```
It turns out only `alibuild_helpers.build.doBuild` additionally calls `alibuild_helpers.utilities.prunePaths`, which clears any environment variable ending in `_VERSION` (with the `ALIBUILD_VERSION` as exception), that also unsets various `*_VERSION` that the Cray PE needs to function:

```
CRAYPE_VERSION
CRAY_LIBSCI_VERSION
DVS_VERSION
GCC_VERSION
GNU_VERSION
INTEL_MAJOR_VERSION
INTEL_MINOR_VERSION
INTEL_VERSION
LIBSCI_VERSION
```

dberzano:
Fixes #530.

### DIFF
--- a/alibuild_helpers/utilities.py
+++ b/alibuild_helpers/utilities.py
@@ -42,6 +42,9 @@ def normalise_multiple_options(option, sep=","):
   return [x for x in ",".join(option).split(sep) if x]
 
 def prunePaths(workDir):
+  if not "ALIENVLVL" in os.environ:
+    # Not in an alienv shell: trust current environment
+    return
   for x in ["PATH", "LD_LIBRARY_PATH", "DYLD_LIBRARY_PATH"]:
     if not x in os.environ:
       continue


ktf:
Is that what we really want? Shouldn't messing up `*_VERSION` be forbidden also when not under alienv?

dberzano:
In principle we don't ever want to mess with the user's environment. We only want to clean up `_VERSION` variables to prevent aliBuild misbehaviours when inside an `alienv` environment, I see no other case where we should clean up variables ending with `_VERSION` (which is a quite generic suffix).

dberzano:
Yes but you realize that this implies rewriting the entirety of our recipes? This is absolutely not what I want to do! :-)

ktf:
Yes, but this change will open a potential can of worms for currently working setup. We added the pruning of version because we did have the problem I indicate above, IIRC.

Alternatively we could change recipes to use `<SOMETHING>_REVISION` to detect alibuild built packages, but that also implies changing recipes (albeit in a compatible way).

Barthelemy:
should we close this PR ?

ktf:
Yes, given we are finally switching to `_REVISION`.

PR #535, opened 2018-04-25
Improve error message when unable to change dir

ktf:
This should improve ability to debug errors people are having
sporadically  when on AFS.


ktf:
Cannot be checked easily.

PR #543, opened 2018-06-20
build: Fix handling of command-line environment variables

akubera:
If environment variables were specified in the command line via `-e` arguments, the build script would try to split the value on `=`, but erroneously passed the string `'1'` (not the integer `1`) into `str.split`.

This fix uses `str.partition` to split on 0 or 1 instances of '=' (the zero case will store the empty string into the environment variable)


akubera:
I'll change it if that's your preferred solution. I was going for unittest-ability and potential reuse by refactoring into a function.

ktf:
I think the partitioning itself is trivial and does not need a unit test as such, what we could probably use is an integration test where the condition which triggered the error is tested on the whole build.

PR #547, opened 2018-08-06
Add pylint to Travis tests

ktf:
Current configuration is the minimal one to have everything green
we should little by little make sure we enable the tests as it
makes sense and as we find the time to actually fix them.


ktf:
Only affects continuous integration which is green. We will enable further checkers as we clean them up.

PR #548, opened 2018-08-09
[WIP] Improve usage of read-only HTTP cache

dberzano:
Allows users outside the build cluster to benefit from precompiled packages

**TODO**
- [ ] Verify Travis/Coverage
- [ ] Add default URL if using `--no-system`


dberzano:
@ktf ready for review.

.pylintrc
```diff
@@ -7,6 +7,9 @@ function-naming-style=camelCase
...
+disable=bad-continuation,bad-whitespace,import-error,ungrouped-imports,missing-docstring,invalid-name,multiple-imports,unidiomatic-typecheck,line-too-long,unused-variable,too-few-public-methods,bare-except,trailing-whitespace,too-many-locals,too-many-branches,too-many-statements,unused-argument,fixme,redefined-builtin,consider-using-ternary,no-else-return,superfluous-parens,too-many-arguments,old-style-class,bad-indentation,wrong-import-order,unused-import,logging-not-lazy,anomalous-backslash-in-string,no-self-use,trailing-newlines,expression-not-assigned,pointless-statement,inconsistent-return-statements,attribute-defined-outside-init,broad-except, deprecated-lambda, duplicate-code, len-as-condition, multiple-statements, no-init, no-value-for-parameter, protected-access,redefined-outer-name, singleton-comparison,super-init-not-called,too-many-lines,too-many-nested-blocks,too-many-return-statements,unneeded-not,too-many-ancestors,deprecated-method,useless-import-alias,useless-object-inheritance,consider-using-set-comprehension,undefined-variable,useless-import-alias,too-many-ancestors,consider-using-dict-comprehension,invalid-envvar-default,logging-format-interpolation
```
ktf: Is the masking due to changes of this patch, or changes in pylint (i.e. is it a new feature?). If the latter, can we keep it clean? The motivation (https://docs.python.org/3/howto/logging.html#optimization) seems to be reasonable (in particular for the `debug` log level).

dberzano:
@ktf I have actually tried the 404 (or any other non-200 and non-300 case) and aliBuild just builds it if it cannot download it. So, no broken installation 🙂

.pylintrc
dberzano: It's a matter of policy, I can change it. I personally find that using `%s`, etc. less readable - I am aware of the motivation but in my case there was nothing performance critical. Let's say, I have favoured readability over performance. But again, your call - I can reformat the messages with `%s` and remove the linter exception.

(Maybe, Python 3's [f-strings](https://realpython.com/python-f-strings/) will address both the performance and readability issue?)

alibuild_helpers/build.py
```diff
@@ -72,37 +71,84 @@ def syncToLocal(self, p, spec):
...
     self.writeStore = ""
     self.architecture = architecture
     self.workdir = workdir
     self.insecure = insecure
+    self.httpTimeoutSec = 15
+    self.httpConnRetries = 4
+    self.httpBackoff = 0.4
+    self.doneOrFailed = []
+
+  def getRetry(self, url, dest=None):
```
ktf: Why not using `curl` `--retry*` options? It's there since `7.12` and even `slc6` has `7.19`.

dberzano: Yes, I am aware - we're using those options already in the publisher scripts for instance. We could use `--retry <nretries> --max-time <timeout> --retry-delay <backoff>`. I don't have a strong opinion, the only point being that `curl` does not support either fractions of seconds in backoff or an incremental backoff.

.pylintrc
ktf: If this is the only patch that triggers it and this is the only place were we use the other syntax, yes, I think we should stay consistent with the rest and we should not have yet another linter exception.

We can of course decide that we want to go the other way around, but then we should do so everywhere, not in one single place.

dberzano: Fully agreed :+1:

ktf:
Ok. And if the server goes back to 200, everything is still ok, correct?

dberzano:
Yes absolutely

dberzano:
So, two more todos before proceeding:

- [ ] Roll back the linter exception and use `%s` in logging functions
- [ ] Use native `curl` cmdline options for timeout/retry/backoff

@ktf anything more?

ktf:
Given this works, maybe we should actually do that in a separate "streamlining" PR so that usual suspects can try it in production already.

PR #551, opened 2018-10-14
add common_recipe option to get DRY recipes

aphecetche:


aphecetche:
See https://alice.its.cern.ch/jira/browse/O2-414

alibuild_helpers/build_template.sh
```diff
@@ -105,13 +105,13 @@ cd "$BUILDDIR"
...
-  (set -x; source "$WORK_DIR/SPECS/$ARCHITECTURE/$PKGNAME/$PKGVERSION-$PKGREVISION/$PKGNAME.sh" 2>&1) | tee "$BUILDROOT/log"
+  (set -x; %(common_recipe)s source "$WORK_DIR/SPECS/$ARCHITECTURE/$PKGNAME/$PKGVERSION-$PKGREVISION/$PKGNAME.sh" 2>&1) | tee "$BUILDROOT/log"
 elif [[ "$CACHED_TARBALL" == "" && $INCREMENTAL_BUILD_HASH != "0" && -f "$BUILDDIR/.build_succeeded" ]]; then
   set -o pipefail
-  (%(incremental_recipe)s) 2>&1 | tee "$BUILDROOT/log"
+  (%(common_recipe)s %(incremental_recipe)s) 2>&1 | tee "$BUILDROOT/log"
 elif [[ "$CACHED_TARBALL" == "" ]]; then
   set -o pipefail
-  (set -x; source "$WORK_DIR/SPECS/$ARCHITECTURE/$PKGNAME/$PKGVERSION-$PKGREVISION/$PKGNAME.sh" 2>&1) | tee "$BUILDROOT/log"
+  (set -x; %(common_recipe)s source "$WORK_DIR/SPECS/$ARCHITECTURE/$PKGNAME/$PKGVERSION-$PKGREVISION/$PKGNAME.sh" 2>&1) | tee "$BUILDROOT/log"
```
ktf: shouldn't you always add a "\n" as well?

aphecetche:
@dberzano Indeed, it does work and ease the work of the recipe authors.

ktf:
ah, right... Another problem is that this will require a new version of aliBuild for any recipe requiring `common_recipe`, which probably means a mass migration to the foreseen `aliBuild` 1.6 for the whole collaboration so it should be announced accordingly.

ktf:
@aphecetche do you plan to continue with this, adding the common recipe to the hash? You should also make sure that if no common recipe is provided, alibuild will not rebuild a package which was built with an old version.

aphecetche:
@ktf unfortunately most probably not in the near future, no. 
I would have to get back into alibuild (as I think I forgot everything about it since that commit...) and that's not a context switch I'm willing to do right now (would rather finish the cmake business before I break for the summer holidays...).

ktf:
Fair enough. I am closing this, then, feel free to reopen when needed, of course.

PR #557, opened 2018-12-04
Set aliDeps temp file mode to text

### LINKED ISSUE #556: aliBuild deps can't write dot file when using python3
When using aliBuild via python3 the deps command can't write the dot file:

```Traceback (most recent call last):
  File "/home/ol11/.local/bin/aliBuild", line 132, in <module>
    doMain(args, parser)
  File "/home/ol11/.local/bin/aliBuild", line 68, in doMain
    disable=args.disable) and sys.exit(0) or sys.exit(1)
  File "/home/ol11/.local/lib/python3.7/site-packages/alibuild_helpers/deps.py", line 60, in deps
    fp.write(result)
  File "/usr/lib64/python3.7/tempfile.py", line 481, in func_wrapper
    return func(*args, **kwargs)
TypeError: a bytes-like object is required, not 'str'
```

The issue is found in the following lines of `alibuild_helpers/deps.py`:

```python
  if args.outdot:
    fp = open(args.outdot, "w")
  else:
    fp = NamedTemporaryFile(delete=False)
  fp.write(dot)
  fp.close()
```

Everything works fine with python2, the error seems to be due to the string/byte-array changes between python2 and python3.

I will *try* to fix this in a way that works for both python versions and open a pull request.

dberzano:
Python 3 writes bytes, not strings. Writing pure text allows us to pass strings
to `write()` as well.

Fixes #556.

### DIFF
--- a/alibuild_helpers/deps.py
+++ b/alibuild_helpers/deps.py
@@ -87,9 +87,9 @@ def doDeps(args, parser):
   dot += "}\n"
 
   if args.outdot:
-    fp = open(args.outdot, "w")
+    fp = open(args.outdot, "wt")
   else:
-    fp = NamedTemporaryFile(delete=False)
+    fp = NamedTemporaryFile(delete=False, mode="wt")
   fp.write(dot)
   fp.close()
 
--- a/tests/test_deps.py
+++ b/tests/test_deps.py
@@ -75,5 +75,9 @@ class DepsTestCase(unittest.TestCase):
 
     doDeps(args, MagicMock())
 
+    # Same check without explicit intermediate dotfile
+    args.outdot = None
+    doDeps(args, MagicMock())
+
 if __name__ == '__main__':
   unittest.main()


olantwin:
I can confirm that this now works with python3 as well as python2.

dberzano:
I have to fix our pedantic coverage check 😆

dberzano:
@ktf all good from the codecov perspective

alibuild_helpers/deps.py
```diff
@@ -87,9 +87,9 @@ def doDeps(args, parser):
   dot += "}\n"
 
   if args.outdot:
-    fp = open(args.outdot, "w")
+    fp = open(args.outdot, "wt")
```
ktf: why is this needed? Isn't "w" == "wt"?

dberzano: That's what I thought. [However](https://stackoverflow.com/questions/23051062/open-files-in-rt-and-wt-modes#comment35854536_23051114):

> In Python 3 there's an additional difference between text and binary file modes (on all platforms). In text mode, read returns Unicode strings. In binary mode, read returns a bytes instance. If you want to write Python 2 code with forwards compatibility in mind, you can use io.open rather than the standard open to get the Python 3 behavior (with unicode versus str instances).

which is our case precisely 😄

PR #562, opened 2019-04-03
Add support for direnv

ktf:


dberzano:
Excellent! Let me test it.

ktf:
Ok, merge at your pleasure. I think we reached consensus this is harmless in the worst case.

PR #576, opened 2019-08-06
Add option to avoid rebuilding when development dependencies change

ktf:


davidrohr:
@ktf : So this PR didn't do what I want, and I took me quite a while to get behind it.
So long story short, the functionality is already there and this PR is not needed.

With AliRoot and O2: alibuild correctly does an incremental build, but it seems to be some kind of CMake issue, where CMake just rebuilds whole O2. But this is not an aliBuild problem, aliBuild does an incremental build as it should.

With FairRoot (and formerly AliTPCCommon): FairRoot and AliTPCCommon do not have an incremental build recipe, so aliBuild will always do a full rebuild for them. Then, they are installed into a different directory. And then, any dependency will also have to do a full rebuild. Don't know if we can (or want to) implement some workaround here. Easiest is probably to just add incremental builds for common development packages.

PR #578, opened 2019-09-06
Ensure that alibuild builds the right tag if folder is already present

davidrohr:


davidrohr:
@ktf @MichaelLettrich : What is the reason travis fails?

alibuild_helpers/build_template.sh
```diff
@@ -75,13 +75,19 @@ fi
...
+    # In case there is a stale link / file, for whatever reason.
+    rm -rf $SOURCEDIR
+    git clone ${GIT_REFERENCE:+--reference $GIT_REFERENCE} "$SOURCE0" "$SOURCEDIR"
+    cd $SOURCEDIR
+    git remote set-url --push origin $WRITE_REPO
+    git checkout "${GIT_TAG}"
+  else
+    # Folder is already present, but check that it is the right tag
+    cd $SOURCEDIR
+    git checkout "${GIT_TAG}"
```
ktf: I think doing this will break incremental builds.

davidrohr:
Hm, incremental builds are only for development packeges, aren't they? Is this part actually run in that case?

ktf:
not sure. the travis error is a unit test which breaks, AFAICT.

davidrohr:
@ktf : Travis errors seem unrelated. An empty PR (#579) yields the same travis errors, so something is wrong with travis for alibuild in general.

I locally tried a development repo and incremental build for O2 and it works.
It also seems the code path I changed is not executed for development repos at all.
And as far as I understand incremental builds will only run for development repos.
So I do not understand what could be broken by my PR.

PR #582, opened 2019-09-30
Force fetch mirrored tags / branches

ktf:
The idea is that if someone moves a tag upstream, we should
still be able to fetch the new one and rebuild. This should be
transparently handled by aliBuild in any case, since alibuild
resolves the commit hash to create the packaging hash.


ktf:
@MichaelLettrich @davidrohr @Barthelemy

Barthelemy:
makes sense

PR #592, opened 2019-11-27
Add timestamp to log when in debug mode

ktf:


ktf:
@davidrohr could you take a look if this would work for you?

ktf:
(only available when --debug is passed).

PR #593, opened 2019-12-14
WIP: Don't MIRROR sources if there is a prebuilt tar in remote store

anerokhi:
So, this stops aliBuild from MIRRORing sources (at least for me). Basically I continued to violate the DRY principle and now the python code not only creates symlinks but does almost everything what `build_template.sh:# Unpack, and relocate` does: https://github.com/alisw/alibuild/blob/930f10efa9bd253f44e7f44bcfc54d4001c8e467/alibuild_helpers/build_template.sh#L263-L274
WIP because I suspect this won't work with Docker. I would like to get some feedback on the direction before trying to make this work in Docker. @sawenzel, @ktf, @dberzano ?

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -464,7 +464,8 @@ def doBuild(args, parser):
 
   # Clone/update repos
   for p in [p for p in buildOrder if "source" in specs[p]]:
-    updateReferenceRepoSpec(args.referenceSources, p, specs[p], args.fetchRepos)
+    if not args.remoteStore or args.forceCloneRepos:
+      updateReferenceRepoSpec(args.referenceSources, p, specs[p], args.fetchRepos)
 
     # Retrieve git heads
     cmd = "git ls-remote --heads %s" % (specs[p].get("reference", specs[p]["source"]))
@@ -768,8 +769,29 @@ def doBuild(args, parser):
                         a=args.architecture,
                         p=spec["package"])
 
+          getstatusoutput('cd {}; {} -xzf {}'.format(workDir, tar(), d))
           getstatusoutput("ln -snf %s %s" % (src, dst1))
           getstatusoutput("ln -snf %s %s" % (src, dst2))
+
+          if args.docker:
+            dockerWrapper = '''
+              docker run --rm -it
+              --user $(id -u):$(id -g)
+              -v {workdir}:/sw
+              -e WORK_DIR=/sw
+              -w /sw
+              {image}
+              bash -ex /sw/{architecture}/{p}/latest
+              '''.replace('\n', ' ').format(
+                image=dockerImage,
+                workdir=workDir,
+                architecture=args.architecture,
+                p=spec["package"],
+              )
+            execute(dockerWrapper)
+          else:
+            getstatusoutput('cd {0}; export WORK_DIR={0}; bash -ex "{1}/relocate-me.sh"'.format(workDir, realpath(dst2)))
+
           info("Using cached build for %s" % p)
         break
       else:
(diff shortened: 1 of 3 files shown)


alibuild_helpers/build.py
```diff
@@ -405,10 +405,11 @@ def doBuild(args, parser):
 
   # Clone/update repos
   for p in [p for p in buildOrder if "source" in specs[p]]:
-    updateReferenceRepoSpec(args.referenceSources, p, specs[p], args.fetchRepos)
+    if not args.remoteStore or args.forceCloneRepos:
+      updateReferenceRepoSpec(args.referenceSources, p, specs[p], args.fetchRepos)
 
     # Retrieve git heads
-    cmd = "git ls-remote --heads %s" % (specs[p]["reference"] if "reference" in specs[p] else specs[p]["source"])
+    cmd = "git ls-remote --heads %s" % (specs[p].get("reference", specs[p]["source"]))
```
ktf: This line we can merge immediately if you have it in a separate PR.

alibuild_helpers/build.py
```diff
@@ -709,8 +710,13 @@ def doBuild(args, parser):
                         a=args.architecture,
                         p=spec["package"])
 
+          getstatusoutput('cd {}; {} -xzf {}'.format(workDir, tar(), d))
           getstatusoutput("ln -snf %s %s" % (src, dst1))
           getstatusoutput("ln -snf %s %s" % (src, dst2))
+
+          os.environ['WORK_DIR'] = workDir
```
ktf: This will set WORK_DIR for the whole script, which we do not want. Can you do it inside the getstatusoutput?

ktf:
How do you determine which tarball to download in the first place if you do not have the sources?

alibuild_helpers/build.py
anerokhi: https://github.com/alisw/alibuild/pull/594

anerokhi:
> How do you determine which tarball to download in the first place if you do not have the sources?

You mean that the hash of `HEAD` participates in package hash calculation? It is not necessary to mirror sources to get it. Even currently the hash(es) may be fetched remotely, see https://github.com/alisw/alibuild/blob/930f10efa9bd253f44e7f44bcfc54d4001c8e467/alibuild_helpers/build.py#L411-L418

When building using the `dontclone` branch of `aliBuild`, I get the same package hashes as with the `master` branch.

ktf:
Ok, I think I understand the logic, but why do you need the extra commands to unpack the tarball and relocate?

languagelawyer:
> why do you need the extra commands to unpack the tarball and relocate?

Why I'm not relying on `build_template.sh`? Because it clones sources **before** checking for cached tarball:
https://github.com/alisw/alibuild/blob/0aced60358294740d6e107a24e2720c9b01e81f8/alibuild_helpers/build_template.sh#L79-L94
https://github.com/alisw/alibuild/blob/0aced60358294740d6e107a24e2720c9b01e81f8/alibuild_helpers/build_template.sh#L113-L136

anerokhi:
> why do you need the extra commands to unpack the tarball and relocate?

Why I'm not relying on `build_template.sh`? Because it clones sources **before** checking for cached tarball:
https://github.com/alisw/alibuild/blob/0aced60358294740d6e107a24e2720c9b01e81f8/alibuild_helpers/build_template.sh#L79-L94
https://github.com/alisw/alibuild/blob/0aced60358294740d6e107a24e2720c9b01e81f8/alibuild_helpers/build_template.sh#L113-L136

anerokhi:
Added code for Docker. Looks ugly, but I don't see how to implement it without code duplication without huge refactoring/rewrite of the whole function.

anerokhi:
@ktf is it necessary to mirror all sources in one step before trying to build packages?

ktf:
There was some rationale behind, but I do not remember. Why?

anerokhi:
The rationale is in #496.

ktf:
Unfortunately there is now a conflict with the parallel fetching of the repos. If you are still interested, can you rebase?

ktf:
Closing this as I do not have a path forward for this. Feel free to reopen if you have a suggestion.

PR #595, opened 2019-12-29
Document the fact we now have a brew package for alibuild

ktf:


ktf:
@mpuccio, you might want to consider this as a better install option on mac in the alice-doc tutorial.

PR #609, opened 2020-03-30
aliBuild/docs - reference section, TOC

maireiphc:
Here are several documentation points to be developed about aliBuild.
The idea would be to have an ~advanced documentation to be used as reference.

Proposed items (@qgp, @marcovanleeuwen) : 
- advanced concepts of aliBuild (e.g. relocation step is little known) and recipes incl. conventions (e.g. variables to export in module file)
- known limitations/issues (e.g. fastjet-config until recently), to simply know what is not expected to work (and possibly decide to fix it)
- basic strategy for comment to be enforced into recipes

(CC to @preghenella)

### DIFF
--- a/docs/reference.markdown
+++ b/docs/reference.markdown
@@ -4,7 +4,14 @@ subtitle: Recipe reference manual
 layout: main
 ---
 
-## Recipes format
+1. [Recipe formats](#Recipes)
+    1. [The header](#TheHeader)
+    2. [The body](#TheBody)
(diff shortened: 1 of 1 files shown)


ktf:
I've improved a bit the relocation description. Can you have a look and eventually comment? Notice that Github generates for your automatically anchors-using-the-title-lowercase-with-dashes.

ktf:
I think this is a good start. Let me know if you have more questions, using the github code comments, otherwise I think we can merge it.

PR #615, opened 2020-04-06
Fix architecture detection when run from git (O2-1308)

ktf:


ktf:
@sawenzel @pzhristov can you try this?

pzhristov:
It solves the problem for me.

PR #621, opened 2020-07-16
Force checkout

ktf:


ktf:
@sawenzel can you check if this solves the issue on AliBI?

PR #627, opened 2020-10-10
Always recreate symlinks to the development package builds.

vkucera:
Currently, symlinks to the builds of dependencies are recreated every time `aliBuild build` is executed, but symlinks to the builds of development packages (both in `$WORK_DIR/$ARCHITECTURE` and `$WORK_DIR/BUILD`) are created (inside `build_template.sh`) only when the development package needs to be rebuilt.

This update makes `aliBuild build` recreate the symlinks to the builds of development packages every time as well.

This allows for a thorough cleanup where one can delete all build symlinks, recreate them only for the builds to be kept with `aliBuild build` and let `aliBuild clean` delete all the other builds.


ktf:
Superseded by #639. Thanks! Closing.

PR #630, opened 2020-10-28
Switch to S3 as default slc7 remote

ktf:
Time to say goodbye to alicache.


ktf:
@TimoWilken this plus your sync script seems to work for me. Could you stress test it, e.g. with alidist@O2-v1.3.0?

TimoWilken:
Building O2 using these patches and alidist O2-v1.3.0 works for me -- aliBuild downloads packages from S3.

PR #635, opened 2020-11-16
Show debug output from s3cmd

TimoWilken:
This makes debugging problems with S3 uploads much easier.

This output is only shown when `--debug` is given.


ktf:
Could you check on top of #634? With that I was able to publish to the ubuntu S3 space.

ktf:
I guess we can close this?

TimoWilken:
Yeah, that's fine!

PR #638, opened 2020-12-15
HTTP backend: create dangling symlinks

TimoWilken:
This is required so revision numbers don't conflict.

Adds a mode to `getRetry` to return the contents of a remote file, to make symlink creation easier.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -91,7 +91,7 @@ class HttpRemoteSync:
     self.httpBackoff = 0.4
     self.doneOrFailed = []
 
-  def getRetry(self, url, dest=None):
+  def getRetry(self, url, dest=None, returnResult=False):
     for i in range(0, self.httpConnRetries):
       if i > 0:
         pauseSec = self.httpBackoff * (2 ** (i - 1))
@@ -99,28 +99,39 @@ class HttpRemoteSync:
         time.sleep(pauseSec)
       try:
         debug("GET %s: processing (attempt %d/%d)" % (url, i+1, self.httpConnRetries))
-        if dest:
-          # Destination specified. Use requests in stream mode
+        if dest or returnResult:
+          # Destination specified -- file (dest) or buffer (returnResult).
+          # Use requests in stream mode
           resp = get(url, stream=True, verify=not self.insecure, timeout=self.httpTimeoutSec)
           size = int(resp.headers.get("content-length", "-1"))
           downloaded = 0
           reportTime = time.time()
-          with open(dest+".tmp", "wb") as destFp:
-            for chunk in resp.iter_content(chunk_size=32768):
-              if chunk:
+          result = []
+
+          try:
+            destFp = open(dest+".tmp", "wb") if dest else None
+            for chunk in filter(bool, resp.iter_content(chunk_size=32768)):
+              if destFp:
                 destFp.write(chunk)
-                downloaded += len(chunk)
-                if size != -1:
-                  now = time.time()
-                  if downloaded == size:
-                    debug("Download complete")
-                  elif now - reportTime > 3:
-                    debug("%.0f%% downloaded..." % (100*downloaded/size))
-                    reportTime = now
-          if size not in [ downloaded, -1 ]:
+              if returnResult:
+                result.append(chunk)
+              downloaded += len(chunk)
+              if size != -1:
+                now = time.time()
+                if downloaded == size:
+                  debug("Download complete")
+                elif now - reportTime > 3:
+                  debug("%.0f%% downloaded..." % (100*downloaded/size))
+                  reportTime = now
+          finally:
+            if destFp:
+              destFp.close()
+
+          if size not in (downloaded, -1):
             raise PartialDownloadError(downloaded, size)
-          os.rename(dest+".tmp", dest)  # we should not have errors here
-          return True
+          if dest:
+            os.rename(dest+".tmp", dest)  # we should not have errors here
+          return b''.join(result) if returnResult else True
         else:
           # For CERN S3 we need to construct the JSON ourself...
           s3Request = re.match("https://s3.cern.ch/swift/v1[/]+([^/]*)/(.*)$", url)
@@ -174,11 +185,7 @@ class HttpRemoteSync:
       self.doneOrFailed.append(spec["hash"])
       return
 
-    cmd = format("mkdir -p %(hd)s && "
-                 "mkdir -p %(ld)s",
-                 hd=spec["tarballHashDir"],
-                 ld=spec["tarballLinkDir"])
-    execute(cmd)
+    execute(" ".join(("mkdir", "-p", spec["tarballHashDir"], spec["tarballLinkDir"])))
     hashList = [x["name"] for x in hashList]
 
     hasErr = False
@@ -187,9 +194,7 @@ class HttpRemoteSync:
       if os.path.isfile(destPath):
         # Do not redownload twice
         continue
-      srcUrl = "{remoteStore}/{storePath}/{pkg}".format(
-        remoteStore=self.remoteStore, storePath=spec["storePath"], pkg=pkg)
-      if self.getRetry(srcUrl, destPath) != True:
+      if not self.getRetry("/".join((self.remoteStore, spec["storePath"], pkg)), destPath):
         hasErr = True
 
     for pkg in pkgList:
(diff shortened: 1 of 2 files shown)


TimoWilken:
I'm building O2 using aliBuild with this PR applied. Symlinks are created correctly:

```bash
$ # for example
$ ls -l sw/TARS/slc7_x86-64/FLUKA/
total 36
lrwxrwxrwx 1 root root   9 Dec 15 15:18 FLUKA -> NoSuchKey
# the following two symlink targets do not exist
lrwxrwxrwx 1 root root 108 Dec 15 15:18 FLUKA-2011-3.0-vmc2-1.slc7_x86-64.tar.gz -> ../../slc7_x86-64/store/e0/e0e01a86eed1a3dc1380b3258ba7ba4631fccbc2/FLUKA-2011-3.0-vmc2-1.slc7_x86-64.tar.gz
lrwxrwxrwx 1 root root 108 Dec 15 15:18 FLUKA-2011-3.0-vmc2-2.slc7_x86-64.tar.gz -> ../../slc7_x86-64/store/e6/e682c4f412c84a15fd49b3311b2aeec940a4f2bf/FLUKA-2011-3.0-vmc2-2.slc7_x86-64.tar.gz
# the following one exists
lrwxrwxrwx 1 root root 108 Dec 15 15:18 FLUKA-2011-3.0-vmc3-1.slc7_x86-64.tar.gz -> ../../slc7_x86-64/store/89/89c0c5c50f645c84548c43212cf644a7b9e78913/FLUKA-2011-3.0-vmc3-1.slc7_x86-64.tar.gz
lrwxrwxrwx 1 root root   9 Dec 15 15:18 FLUKA_VMC -> NoSuchKey
lrwxrwxrwx 1 root root   9 Dec 15 15:18 FLUKA_VMC-2011-3.0-vmc3-1.slc7_x86-64.tar.gz -> NoSuchKey
lrwxrwxrwx 1 root root   9 Dec 15 15:18 FLUKA_VMC-2011-3.0-vmc3-2.slc7_x86-64.tar.gz -> NoSuchKey
lrwxrwxrwx 1 root root   9 Dec 15 15:18 FLUKA_VMC-2011-3.0-vmc3-3.slc7_x86-64.tar.gz -> NoSuchKey
lrwxrwxrwx 1 root root   9 Dec 15 15:18 FLUKA_VMC-2011-3.0-vmc4-1.slc7_x86-64.tar.gz -> NoSuchKey
```

Nonexistent symlinks are linked to `NoSuchKey` instead of `unknown`. Symlinks from S3 now point to their targets (even if those targets don't exist on disk) instead of `unknown`.

@ktf, fine to merge in my opinion, unless you'd like me to change anything!

alibuild_helpers/build.py
```diff
@@ -174,11 +185,7 @@ def syncToLocal(self, p, spec):
       self.doneOrFailed.append(spec["hash"])
       return
 
-    cmd = format("mkdir -p %(hd)s && "
-                 "mkdir -p %(ld)s",
-                 hd=spec["tarballHashDir"],
-                 ld=spec["tarballLinkDir"])
-    execute(cmd)
+    execute(" ".join(("mkdir", "-p", spec["tarballHashDir"], spec["tarballLinkDir"])))
```
ktf, line 188: why not simply `format("mkdir -p %s %s" % (spec["tarballHashDir"], spec["tarballLinkDir"]))`?

alibuild_helpers/build.py
```diff
@@ -91,36 +91,47 @@ def __init__(self, remoteStore, architecture, workdir, insecure):
     self.httpBackoff = 0.4
     self.doneOrFailed = []
 
-  def getRetry(self, url, dest=None):
+  def getRetry(self, url, dest=None, returnResult=False):
```
ktf, line 94: This is fine for now, however we should probably simply rely on requests rather than implement our own logic.

alibuild_helpers/build.py
TimoWilken: OK, I've changed it!

TimoWilken:
The failing test is an argument parsing one. Specifically, the following args:
```
build zlib --architecture ubuntu1804_x86-64
```
the test expects `https://s3.cern.ch/swift/v1/alibuild-repo` as the remote store and `noSystem == True`, which would be set if the remote store was set correctly.

Looking at the argument parsing code, the remote store is only set to the expected default [if `args.architecture` is `slc7_x86-64`][if] -- so this fails if the architecture is `ubuntu1804_x86-64`, as in the failing test.

@ktf: should I change the test or the argument parsing code?

[if]: https://github.com/alisw/alibuild/blob/master/alibuild_helpers/args.py#L234

ktf:
I guess the correct thing to do would be to add ubuntu to use the remote store. We are uploading to s3 for ubuntu, right?

TimoWilken:
@ktf yes, the Ubuntu builder uploads to S3 (see packages at <https://s3.cern.ch/swift/v1/alibuild-repo/?prefix=TARS/ubuntu>). I've added it to the "approved" architectures.

ktf:
Do you understand the other errors?

TimoWilken:
@ktf no, I don't.
```
+ coverage run -a alibuild/aliBuild --aggressive-cleanup --docker -a slc7_x86-64 --always-prefer-system -d build zlib
DEBUG: Analytics previously disabled.
DEBUG: Current working directory is /home/runner/work/alibuild
DEBUG: Package RooUnfold has been disabled by current default.
DEBUG: Package treelite has been disabled by current default.
DEBUG: Building for architecture slc7_x86-64
DEBUG: Number of parallel builds: 2
DEBUG: Using aliBuild from alibuild@WARNING: Package(s) not found: alibuild recipes in alidist@573358b87a1d1db288811fa4311ace1c9438c9e7
DEBUG: Overrides for package defaults-release: {'env': OrderedDict([('CXXFLAGS', '-fPIC -g -O2 -std=c++11'), ('CFLAGS', '-fPIC -g -O2'), ('CMAKE_BUILD_TYPE', 'RELWITHDEBINFO')])}
ERROR: The following packages are system requirements and could not be found:
ERROR: 
ERROR: - make
ERROR: - termcap
ERROR: - yacc-like
ERROR: 
ERROR: Please run:
ERROR: 
ERROR: 	aliDoctor zlib
ERROR: 
ERROR: to get a full diagnosis.
```

`pythonpackage.yml` already installs `bison` and `flex` (which `yacc-like.sh` checks for), `build-essential` (which depends on `make`), and `libncurses-dev` (needed for `termcap`). I don't know why they aren't picked up.

TimoWilken:
@ktf I didn't realise it was trying to use docker initially. I get the same error locally, maybe something is wrong with the slc7-builder image? `aliDoctor --docker -a slc7_x86-64 --always-prefer-system -d zlib` shows:
```
DEBUG: Package yacc-like can only be managed via alibuild.
ERROR: Package yacc-like is a system requirement and cannot be found.
ERROR: This is due to the fact that the following script fails:
ERROR: 
ERROR: brew() { true; }; which bison && which flex
ERROR: with the following output:
ERROR: 
ERROR: yacc-like: /usr/bin/bash: /usr/bin/bash: cannot execute binary file
ERROR: yacc-like: 
ERROR: Please install bison and flex develpment package on your system.
ERROR: If they are there, make sure you have them in your default path or check you have `which` installed.
ERROR: 
ERROR: 
DEBUG: Package make can only be managed via alibuild.
ERROR: Package make is a system requirement and cannot be found.
ERROR: This is due to the fact that the following script fails:
ERROR: 
ERROR: brew() { true; }; type make
ERROR: with the following output:
ERROR: 
ERROR: make: /usr/bin/bash: /usr/bin/bash: cannot execute binary file
ERROR: make: 
ERROR: Please install make package on your system:
ERROR:   * On RHEL-compatible systems: make
ERROR:   * On Ubuntu-compatible systems: make
ERROR: 
ERROR: 
DEBUG: Package termcap can only be managed via alibuild.
ERROR: Package termcap is a system requirement and cannot be found.
ERROR: This is due to the fact that the following script fails:
ERROR: 
ERROR: brew() { true; }; printf "#include <termcap.h>\n" | c++ -xc++ - -c -o /dev/null
ERROR: with the following output:
ERROR: 
ERROR: termcap: /usr/bin/bash: /usr/bin/bash: cannot execute binary file
ERROR: termcap: 
ERROR: Please install the ncurses development package on your system (usually ncurses-devel or libncurses-dev)
ERROR: 
```
I'm investigating, but would appreciate hints.

ktf:
yeah, looks like it. Can you dig a bit further? This is otherwise fine.

TimoWilken:
@ktf I've tracked it down to a wrong entrypoint for the SLC7 container. alisw/docks#204 fixes this, at least on my laptop. I've built the new docker image and pushed it to dockerhub.

PR #639, opened 2020-12-17
Always recreate symlinks to the development package builds.

ktf:


ktf:
Same as #627, just rebased.

PR #640, opened 2020-12-18
If available, have architecture specific overrides

ktf:
This works in parallel of the usual "defaults" mechanism to override recipes on
a per architecture basis. For example to disable a given common package like
autotools or to force some compiler options which are specific for a given
architecture.

### DIFF
--- a/alibuild_helpers/utilities.py
+++ b/alibuild_helpers/utilities.py
@@ -192,7 +192,7 @@ def readDefaults(configDir, defaults, error, architecture):
       error(err)
       sys.exit(1)
     for x in ["env", "disable", "overrides"]:
-      defaultsMeta.get(x, {}).update(archMeta.get(x, {}))
+      defaultsMeta.setdefault(x, {}).update(archMeta.get(x, {}))
     defaultsBody += "\n# Architecture defaults\n" + archBody
   return (defaultsMeta, defaultsBody)
 
--- a/alibuild_helpers/utilities.py
+++ b/alibuild_helpers/utilities.py
@@ -171,7 +171,7 @@ def filterByArchitecture(arch, requires):
     if re.match(matcher, arch):
       yield require
 
-def readDefaults(configDir, defaults, error):
+def readDefaults(configDir, defaults, error, architecture):
   defaultsFilename = "%s/defaults-%s.sh" % (configDir, defaults)
   if not exists(defaultsFilename):
     viableDefaults = ["- " + basename(x).replace("defaults-","").replace(".sh", "")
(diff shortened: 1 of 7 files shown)


alibuild_helpers/utilities.py
```diff
@@ -183,6 +183,17 @@ def readDefaults(configDir, defaults, error):
...
+  archDefaults = "%s/defaults-%s.sh" % (configDir, architecture)
+  archMeta = {}
+  archBody = ""
+  if exists(archDefaults):
+    err, archMeta, archBody = parseRecipe(getRecipeReader(defaultsFilename))
+    if err:
+      error(err)
+      sys.exit(1)
+    for x in ["env", "disable", "overrides"]:
+      defaultsMeta.get(x, {}).update(archMeta.get(x, {}))
```
TimoWilken, line 195: This won't apply arch overrides from an `x` if `defaultsMeta` does not already have an `x` entry. Is that the desired behaviour? If not, the following might be better (works in Python 2 and 3):
```python
defaultsMeta.setdefault(x, {}).update(archMeta.get(x, {}))
```

TimoWilken (changes requested):
This PR breaks unit tests with:
```
ERROR: test_doRealInit (test_init.InitTestCase)
----------------------------------------------------------------------
  File "/opt/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/site-packages/mock/mock.py", line 1346, in patched
    return func(*newargs, **newkeywargs)
  File "/home/runner/work/alibuild/alibuild/tests/test_init.py", line 106, in test_doRealInit
    doInit(args)
  File "/home/runner/work/alibuild/alibuild/alibuild_helpers/init.py", line 55, in doInit
    (err, overrides, taps) = parseDefaults([], defaultsReader, debug)
  File "/home/runner/work/alibuild/alibuild/alibuild_helpers/utilities.py", line 278, in parseDefaults
    defaultsMeta, defaultsBody = defaultsGetter()
  File "/home/runner/work/alibuild/alibuild/alibuild_helpers/init.py", line 54, in <lambda>
    defaultsReader = lambda: readDefaults(args.configDir, args.defaults, error, args.architecture)
AttributeError: 'Namespace' object has no attribute 'architecture'
```

The `Namespace` defined on `test_init.py` line 96 needs an `architecture` entry.

alibuild_helpers/utilities.py
ktf: Nice catch.

ktf:
Thanks for checking. I think I fixed both.

TimoWilken (approved):
Thanks! Fine by me when CI passes.

PR #644, opened 2021-01-13
Optimize symlink creation

ktf:
This seems to cut 4 / 5 seconds from an empty build, bringing it
from 17 seconds to 12 in the O2 case, at least on my machine.


TimoWilken (approved):
Looks good to me!

PR #652, opened 2021-02-12
Also restore leading ../../ for HTTPS remote

TimoWilken:
This prefix is removed in `S3RemoteStore.syncToRemote`, but needs to be restored here too, as the HTTPS remote is used to access S3 as well.

@ktf, I assume the HTTPS remote is *only* used to access S3. Is that right? If not, this might need to be handled specially.


ktf:
Yes, I think we deprecated the old ali-ci.cern.ch remote, no?

TimoWilken:
Yes, you're right. And the symlink business is S3-specific already, so this change would only break what would already be broken anyway. :)

PR #657, opened 2021-02-26
Use args of logging functions for lazy formatting

TimoWilken:
Logging functions (like `debug`, `info`, `error`) can take extra arguments, which are substituted into the message using printf-like syntax (`%s` and the like). This formatting is only done if the given logging level is actually enabled, giving some minor speed up in case debug logging is disabled. It also makes the code cleaner, as e.g. `debug(format("%s %s", a, b))` can be written as `debug("%s %s", a, b)` instead.

This patch also updates unit tests to work with the built-in logging convention.

Also includes fixes for the minor problem where `err = execute(cmd); debug(err, cmd)` would not output `cmd`; only `err` (the exit code) without explanation. I've changed this to output both `cmd` and `err` with a more descriptive message.

Tested and working.

### DIFF
--- a/tests/test_parseRecipe.py
+++ b/tests/test_parseRecipe.py
@@ -57,8 +57,8 @@ found unexpected end of stream
 class Recoder(object):
   def __init__(self):
     self.buffer = ""
-  def __call__(self, s):
-    self.buffer += s
+  def __call__(self, s, *a):
+    self.buffer += s % a
 
 class BufferReader(object):
   def __init__(self, filename, recipe):
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -95,7 +95,7 @@ class HttpRemoteSync:
     for i in range(0, self.httpConnRetries):
       if i > 0:
         pauseSec = self.httpBackoff * (2 ** (i - 1))
-        debug("GET %s failed: retrying in %.2f" % (url, pauseSec))
+        debug("GET %s failed: retrying in %.2f", url, pauseSec)
         time.sleep(pauseSec)
         # If the download has failed, enable debug output, even if it was
         # disabled before. We disable debug output for e.g. symlink downloads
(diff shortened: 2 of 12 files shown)


tests/test_parseRecipe.py
```diff
@@ -57,8 +57,8 @@
 class Recoder(object):
   def __init__(self):
     self.buffer = ""
-  def __call__(self, s):
-    self.buffer += s
+  def __call__(self, s, *a):
+    self.buffer += s % a
```
ktf, line 61: can you do `format(s, **kwds) % *a` instead, so that we can also use the keywords?

ktf (commented):
Maybe we can keep the current keyword based functionality?

tests/test_parseRecipe.py
TimoWilken: I'm not a big fan of overriding the builtin `format` (which does something else), though `s % (kwds or a)` would work, I suppose...

The problem is that the built-in logger doesn't have that functionality, I think. It uses keywords for special situations (e.g. to tell it that it should print exception info attached to the log line). There might be a way to retrofit that, but it's likely a bit hacky, if not impossible.

ktf: For the record, the reason why we have an override for the format method is that at the time we had to support python 2.5. If you want, feel free to replace all of them with `str.format`. That said, please, do not go and replace the current "keyword based" solution everywhere with positional formatting.

Given you replaced the keyword functionality only in an handful of places let's merge this.

ktf (approved):
For the record, the reason why we have an override for the format method is that at the time we had to support python 2.5. If you want, feel free to replace all of them with str.format. That said, please, do not go and replace the current "keyword based" solution everywhere with positional formatting.

Given you replaced the keyword functionality only in an handful of places let's merge this.

PR #658, opened 2021-02-26
Separate local-only builds from remote revisions

TimoWilken:
This prevents interference between the remote store and local builds in the case where another machine builds a package and uploads it to the remote store with a revision number that conflicts with a local package, in case that package wasn't uploaded to the remote store. In that case, local revision numbers become a mess, and packages could start depending on the wrong revision numbers of dependencies.

It also prevents builds intended only for local use from being uploaded to the remote. This is important as local-only builds might otherwise be uploaded to the remote much later, when some of their dependencies might conflict with new ones on the remote, and would be renamed. In that case, we currently don't update the main package, so it would depend on (and e.g. try to source scripts from) an incorrect revision number of its dependency.

With these changes, if running aliBuild with a read-write remote when some local revisions are left over in the local store from a previous run, then those revisions are not uploaded, and may be deleted from the local store. Running aliBuild repeatedly with a read-only remote preserves local revisions.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -67,7 +67,6 @@ def getDirectoryHash(d):
 
 # Helper class which does not do anything to sync
 class NoRemoteSync:
-  writeStore = ""
   def syncToLocal(self, p, spec):
     pass
   def syncToRemote(self, p, spec):
@@ -728,7 +727,7 @@ def doBuild(args, parser):
       logger_handler.setFormatter(
           LogFormatter("%%(asctime)s:%%(levelname)s:%s:%s:%s: %%(message)s" %
                        (mainPackage, p, args.develPrefix if "develPrefix" in args else mainHash[0:8])))
-    if spec["package"] in develPkgs and syncHelper.writeStore:
+    if spec["package"] in develPkgs and getattr(syncHelper, "writeStore", None):
       warning("Disabling remote write store from now since %s is a development package.", spec["package"])
       syncHelper.writeStore = ""
 
@@ -795,16 +794,17 @@ def doBuild(args, parser):
     busyRevisions = set()
     # We can tell that the remote store is read-only if it has an empty or
     # no writeStore property. See below for explanation of why we need this.
-    revisionPrefix = "" if syncHelper.writeStore else "local"
+    revisionPrefix = "" if getattr(syncHelper, "writeStore", "") else "local"
     for d in packages:
-      match = re.match(
-        "../../%s/store/[0-9a-f]{2}/([0-9a-f]+)/%s-%s-((?:local)?[0-9]+).%s.tar.gz$" %
-        tuple(map(re.escape, (args.architecture, spec["package"],
-                              spec["version"], args.architecture))),
-        readlink(d))
-      if not match:
+      realPath = readlink(d)
+      matcher = format("../../%(a)s/store/[0-9a-f]{2}/([0-9a-f]*)/%(p)s-%(v)s-((?:local)?[0-9]*).%(a)s.tar.gz$",
+                       a=args.architecture,
+                       p=spec["package"],
+                       v=spec["version"])
+      m = re.match(matcher, realPath)
+      if not m:
         continue
-      h, revision = match.groups()
+      h, revision = m.groups()
 
       if h != spec["hash"]:
         if revision.startswith(revisionPrefix) and revision[len(revisionPrefix):].isdigit():
@@ -815,7 +815,7 @@ def doBuild(args, parser):
 
       # Don't re-use local revisions when we have a read-write store, so that
       # packages we'll upload later don't depend on local revisions.
-      if syncHelper.writeStore and "local" in revision:
+      if getattr(syncHelper, "writeStore", False) and "local" in revision:
         continue
 
       # If we have an hash match, we use the old revision for the package
@@ -824,14 +824,23 @@ def doBuild(args, parser):
       if spec["package"] in develPkgs and "incremental_recipe" in spec:
         spec["obsolete_tarball"] = d
       else:
-        debug("Package %s with hash %s is already found in %s. Not building.",
-              p, h, d)
-        getstatusoutput("ln -snf %s-%s %s/%s/%s/latest-%s" % (
-          spec["version"], spec["revision"],
-          workDir, args.architecture, spec["package"], spec["build_family"]))
-        getstatusoutput("ln -snf %s-%s %s/%s/%s/latest" % (
-          spec["version"], spec["revision"],
-          workDir, args.architecture, spec["package"]))
+        debug("Package %s with hash %s is already found in %s. Not building.", p, h, d)
+        src = format("%(v)s-%(r)s",
+                     w=workDir,
+                     v=spec["version"],
+                     r=spec["revision"])
+        dst1 = format("%(w)s/%(a)s/%(p)s/latest-%(bf)s",
+                      w=workDir,
+                      a=args.architecture,
+                      p=spec["package"],
+                      bf=spec["build_family"])
+        dst2 = format("%(w)s/%(a)s/%(p)s/latest",
+                      w=workDir,
+                      a=args.architecture,
+                      p=spec["package"])
+
+        getstatusoutput("ln -snf %s %s" % (src, dst1))
+        getstatusoutput("ln -snf %s %s" % (src, dst2))
         info("Using cached build for %s", p)
       break
 
@@ -841,8 +850,8 @@ def doBuild(args, parser):
     # somewhere else, the next revision N might be assigned there, and would
     # conflict with our revision N.
     if "revision" not in spec:
-      # The regex finding busyRevisions above already ensures that revision
-      # numbers start with revisionPrefix, and has stripped the prefix.
+      # The code finding busyRevisions above already ensures that revision
+      # numbers start with revisionPrefix, and has left us plain ints.
       spec["revision"] = revisionPrefix + str(
         min(set(range(1, max(busyRevisions) + 2)) - busyRevisions)
         if busyRevisions else 1)
@@ -887,7 +896,6 @@ def doBuild(args, parser):
       if spec["devel_hash"]+spec["deps_hash"] == spec["old_devel_hash"]:
         info("Development package %s does not need rebuild", spec["package"])
         buildOrder.pop(0)
-        packageIterations = 0
         continue
 
     # Now that we have all the information about the package we want to build, let's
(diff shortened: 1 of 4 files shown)


TimoWilken:
TODO: create a new alibuild release and get the CI builders to use it when merged.

TimoWilken:
I think I've figured out why aliBuild with the local-revision patch rebuilt some packages: I built using the HTTPS remote store first, using a clean `sw/` directory. Then I built with no remote store, but only deleting O2 from the store, keeping the rest of `sw/` from the previous run to save time. During the second run, lots of packages suddenly depended on GCC-Toolchain, which changed their hash -- presumably because this package was built during the previous run and was now available in the store, where those packages had used the system GCC before.

TimoWilken:
@ktf: ready to merge in my opinion, I can make a new alibuild release and deploy it for the CI builders gradually.

alibuild_helpers/build.py
```diff
@@ -67,6 +67,7 @@ def getDirectoryHash(d):
 
 # Helper class which does not do anything to sync
 class NoRemoteSync:
+  writeStore = ""
```
ktf, line 70: This is just some cleanup, right? Can you do it in a separate PR?

alibuild_helpers/build.py
```diff
@@ -792,51 +792,60 @@ def downloadTask(p):
...
       else:
-        busyRevisions.append(revision)
-
-    if not "revision" in spec and busyRevisions:
-      spec["revision"] = min(set(range(1, max(busyRevisions)+2)) - set(busyRevisions))
-    elif not "revision" in spec:
-      spec["revision"] = "1"
+        debug("Package %s with hash %s is already found in %s. Not building.",
+              p, h, d)
+        getstatusoutput("ln -snf %s-%s %s/%s/%s/latest-%s" % (
```
ktf, line 829: Mmm... No, please do not change the key based formatting. If you want to avoid the `format` helper, simply use str.format with the actual name.

alibuild_helpers/cmd.py
```diff
@@ -10,10 +10,9 @@
 
 def execute(command, printer=debug):
   popen = subprocess.Popen(command, shell=is_string(command), stdout=subprocess.PIPE)
-  lines_iterator = iter(popen.stdout.readline, "")
-  for line in lines_iterator:
+  for line in iter(popen.stdout.readline, ""):
     if not line: break
-    printer(to_unicode(line).strip("\n")) # yield line
+    printer("%s", to_unicode(line).strip("\n"))
```
ktf, line 15: Why doing the formatting for "%s"?

alibuild_helpers/cmd.py
```diff
@@ -10,10 +10,9 @@
 
 def execute(command, printer=debug):
   popen = subprocess.Popen(command, shell=is_string(command), stdout=subprocess.PIPE)
-  lines_iterator = iter(popen.stdout.readline, "")
-  for line in lines_iterator:
+  for line in iter(popen.stdout.readline, ""):
```
ktf, line 13: Also this is unrelated to the PR. Can you please split it off?

tests/test_cmd.py
```diff
@@ -16,7 +16,7 @@ class CmdTestCase(unittest.TestCase):
     def test_execute(self, mock_debug):
       err = execute("echo foo", mock_debug)
       self.assertEqual(err, 0)
-      self.assertEqual(mock_debug.mock_calls, [call('foo')])
+      self.assertEqual(mock_debug.mock_calls, [call("%s", "foo")])
```
ktf, line 19: Why the extra formatting?

alibuild_helpers/build.py
```diff
@@ -1205,7 +1215,11 @@ def downloadTask(p):
 
     dieOnError(err, buildErrMsg)
 
-    syncHelper.syncToRemote(p, spec)
+    # Make sure not to upload local-only packages! These might have been
+    # produced in a previous run with a read-only remote store.
+    if not spec["revision"].startswith("local"):
```
ktf, line 1220: We should stop at the first local package, no?

ktf (changes requested):

alibuild_helpers/build.py
TimoWilken: Sorry, that's a holdover from some cleanup I was doing before. I'll change that.

alibuild_helpers/cmd.py
TimoWilken: Because `line` can contain `%` signs (e.g. `[100%] Compiling xyz.cxx`), which confuses the string formatting.

tests/test_cmd.py
TimoWilken: This is changed because of the change above. In the general case, an output line can contain `%`, so this needs the extra `%s`.

alibuild_helpers/build.py
TimoWilken: What should happen in the following case:
```
 D-1        E-local1
  |          |
 B-1        C-local1
   \        /
    A-local1
```
(i.e. A depends partly on local and partly on remote packages, but the remote packages are consistent; `buildOrder` is something like `[D, E, B, C, A]`)

I suppose `B-1` and `D-1` should've been uploaded in a previous alibuild run, so it might be alright to stop uploading when we hit `E-local1`. But if that previous upload failed, would we want it to happen now to make local and remote consistent again?

alibuild_helpers/build.py
TimoWilken: I've undone this.

alibuild_helpers/cmd.py
TimoWilken: Undone.

alibuild_helpers/build.py
ktf: I assume that if B-1 dependend on E, it would actually be called B-local1, no? If that is the case, you are right, we should not stop.

TimoWilken: Yes, if B depended on E, it would get a local revision number. When we have a read-only remote, no new packages with a non-local revision are produced, only existing ones reused.

alibuild_helpers/build.py
```diff
@@ -1205,7 +1214,11 @@ def downloadTask(p):
...
+      syncHelper.syncToRemote(p, spec)
```
ktf, line 1220: Can we invert the logic here? Something like:

```
if spec["revision"].startswith("local"):
  continue       # break?
```

alibuild_helpers/build.py
```diff
@@ -878,6 +886,7 @@ def downloadTask(p):
       if spec["devel_hash"]+spec["deps_hash"] == spec["old_devel_hash"]:
         info("Development package %s does not need rebuild", spec["package"])
         buildOrder.pop(0)
+        packageIterations = 0
```
ktf, line 889: What is this for?

alibuild_helpers/build.py
```diff
@@ -792,51 +791,60 @@ def downloadTask(p):
...
-    elif not "revision" in spec:
-      spec["revision"] = "1"
+        debug("Package %s with hash %s is already found in %s. Not building.",
+              p, h, d)
+        getstatusoutput("ln -snf {ver}-{rev} {wd}/{arch}/{pkg}/latest-{family}".format(
+          ver=spec["version"], rev=spec["revision"], wd=workDir,
+          arch=args.architecture, pkg=spec["package"], family=spec["build_family"]))
+        getstatusoutput("ln -snf {ver}-{rev} {wd}/{arch}/{pkg}/latest".format(
+          ver=spec["version"], rev=spec["revision"], wd=workDir,
+          arch=args.architecture, pkg=spec["package"]))
```
ktf, line 833: Did you change this wrt the current code or is it just a nicer version? Can you do a separate PR for this?

alibuild_helpers/build.py
TimoWilken: I think this was forgotten at some point -- this tracks the number of times the current package was rebuilt, but we've just decided we're done with the current package, so `packageIterations` should be reset.

As a side note, this was/is really unlikely to actually cause a bug because the limit of iterations is set at 20, so one or two more for the next package still won't trip the limit. (Which is why it wasn't noticed before, I presume.)

alibuild_helpers/build.py
```diff
@@ -792,51 +791,60 @@ def downloadTask(p):
...
-      realPath = readlink(d)
-      matcher = format("../../%(a)s/store/[0-9a-f]{2}/([0-9a-f]*)/%(p)s-%(v)s-([0-9]*).%(a)s.tar.gz$",
-                       a=args.architecture,
-                       p=spec["package"],
-                       v=spec["version"])
-      m = re.match(matcher, realPath)
-      if not m:
+      match = re.match(
+        "../../%s/store/[0-9a-f]{2}/([0-9a-f]+)/%s-%s-((?:local)?[0-9]+).%s.tar.gz$" %
+        tuple(map(re.escape, (args.architecture, spec["package"],
```
ktf, line 801: `[re.escape(x) for x in [args.architecture, spec["package"], spec["version"], args.architecture]]` but then again, why not simply using `str.format`?

alibuild_helpers/build.py
TimoWilken: No change, just "nicer". I can make a separate PR.

alibuild_helpers/build.py
```diff
@@ -727,7 +727,7 @@ def downloadTask(p):
       logger_handler.setFormatter(
           LogFormatter("%%(asctime)s:%%(levelname)s:%s:%s:%s: %%(message)s" %
                        (mainPackage, p, args.develPrefix if "develPrefix" in args else mainHash[0:8])))
-    if spec["package"] in develPkgs and getattr(syncHelper, "writeStore", None):
+    if spec["package"] in develPkgs and getattr(syncHelper, "writeStore", False):
```
ktf, line 730: Why changing this here? If it is not strictly necessary, can you move it to a separate PR (which can go in immediately)?

TimoWilken: OK, will do.

alibuild_helpers/build.py
ktf: Yes, please. We can merge that PR immediately. Given the PR is about a quite critical change, I would like it to be distilled to its minimum.

TimoWilken:
One minute, I just realised I reverted a bit too much.

alibuild_helpers/build.py
TimoWilken: Done.

alibuild_helpers/build.py
TimoWilken: Reverted or now; I'll include it in the separate cleanup PR.

alibuild_helpers/build.py
TimoWilken: Reverted.

alibuild_helpers/build.py
TimoWilken: Reverted; I'll include it in the cleanup PR.

Issue #660, opened 2021-03-05
Python version for alibuild/master


davidrohr:
Using alibuild/master fails for me with
```
Traceback (most recent call last):
  File "/home/qon/alice/alibuild/aliBuild", line 132, in <module>
    doMain(args, parser)
  File "/home/qon/alice/alibuild/aliBuild", line 82, in doMain
    printer, msg, code = doBuild(args, parser)
  File "/home/qon/alice/alibuild/alibuild_helpers/build.py", line 555, in doBuild
    err = execute(cmd, h)
  File "/home/qon/alice/alibuild/alibuild_helpers/cmd.py", line 16, in execute
    printer("%s", to_unicode(line).strip("\n"))
TypeError: __call__() takes 2 positional arguments but 3 were given
```
while 1.7.3 works well. I am using python 3.7. Any ideas?

ktf:
@TimoWilken can you have a look?

TimoWilken:
@davidrohr @pzhristov It seems to be working locally for me on Python 3.9.2 and 2.7.18. Which packages are you building?

davidrohr:
O2

pzhristov:
I am building AliPhysics

PR #661, opened 2021-03-05
Do not let the printer do the formatting

ktf:


ktf:
@davidrohr @pzhristov can you see if this fixes your issue?

davidrohr:
@ktf : thx, this fixes #660

pzhristov:
Yes, it does, but now I have other problems in the installation of AliPhysics - some libraries from AliRoot are not found. Not sure if it has to do with the same issue.

ktf:
Timo's fix is actually more complete.

PR #662, opened 2021-03-05
Fix logging errors

TimoWilken:
This should catch the last places where various things are used like loggers, but don't support formatting.


ktf (approved):

ktf:
@davidrohr

PR #666, opened 2021-03-11
Calculate separate hashes for local and remote revisions

TimoWilken:
Local revisions get a different hash (with the same information as remote revisions, but an extra "local" string in the input to the hash). This is to avoid conflicts where a remote might gain a package with the same hash as a locally-built package; in that case, the two would conflict. With this patch, the two revisions of the same package should be able to coexist.

Syncing from and to remote stores can always use the remote-specific hashes. Once we have assigned a revision to the package (or reused an existing one), we know which hash (remote or local) to assign and use for the build.

This commit also extracts the hash calculation into a separate function for cleanliness.

The same hashes are still generated for the same specs, so existing tarballs on the remote store are used.

A bit of reorganisation was necessary to move the hash calculation into the main build loop, as we need to know our dependencies' hashes at the time we calculate our own hash, but to know their hashes, they must have first been assigned a revision number (so that we know whether to use the local or remote hash).

I've tested this by building O2 and AliPhysics on slc7; aliBuild successfully reuses existing packages from the store, and builds some locally with localN revisions. I haven't run into a situation where a remote store package was built with the same config as an already-existing local package, but the separate local/remote hashes should solve that. I think the best way to test that is probably to deploy this change on a CI builder for a little while and see how it goes.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -171,58 +171,61 @@ class HttpRemoteSync:
     return None
 
   def syncToLocal(self, p, spec):
-    if spec["hash"] in self.doneOrFailed:
-      debug("Will not redownload %s with build hash %s", p, spec["hash"])
+    if spec["remote_revision_hash"] in self.doneOrFailed:
+      debug("Will not redownload %s with build hash %s",
+            p, spec["remote_revision_hash"])
       return
 
-    debug("Updating remote store for package %s@%s", p, spec["hash"])
-    hashListUrl = format("%(rs)s/%(sp)s/",
-                        rs=self.remoteStore,
-                        sp=spec["storePath"])
-    pkgListUrl = format("%(rs)s/%(sp)s/",
-                        rs=self.remoteStore,
-                        sp=spec["linksPath"])
+    debug("Updating remote store for package %s@%s",
+          p, spec["remote_revision_hash"])
+    hashListUrl = "{rs}/{sp}/".format(rs=self.remoteStore, sp=spec["remote_store_path"])
+    pkgListUrl = "{rs}/{lp}/".format(rs=self.remoteStore, lp=spec["remote_links_path"])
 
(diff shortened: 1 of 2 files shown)


alibuild_helpers/build.py
```diff
@@ -262,8 +265,8 @@ def syncToRemote(self, p, spec):
                  workdir=self.workdir,
                  remoteStore=self.remoteStore,
                  rsyncOptions=self.rsyncOptions,
-                 storePath=spec["storePath"],
-                 linksPath=spec["linksPath"],
+                 storePath=spec["remote_store_path"],
+                 linksPath=spec["remote_links_path"],
```
ktf, line 269: Why do we need to rename those in the same PR?

alibuild_helpers/build.py
```diff
@@ -363,6 +367,72 @@ def createDistLinks(spec, specs, args, repoType, requiresType):
...
+  spec = specs[package]
+
+  if "remote_revision_hash" in spec and "local_revision_hash" in spec:
+    # We've already calculated these hashes before, so no need to do it again.
+    # This also works around a bug, where after the first hash calculation,
+    # some attributes of spec are changed (e.g. append_path and prepend_path
+    # entries are turned from strings into lists), which changes the hash on
+    # subsequent calculations.
+    return
+
```
ktf, line 385: Sorry, but this make it really difficult to understand what changed. Can we have the refactoring separate from the logic change?

ktf (changes requested):
Maybe it's the PR number influence, but as discussed, we should minimize risks and split the PR in three:
One where rename the various fields to something closer to their final name.
We introduce storeHash() using the current logic.
We rework 2. to the final result, with the changed logic.
This way the critical changes are limited to 3 and we have less to review in case of issues. Keep around the current version, so that we can make sure that 1+2+3 == #666

TimoWilken:
Closing in favour of #667.

PR #667, opened 2021-03-11
Calculate separate hashes for local and remote revisions

TimoWilken:
Local revisions get a different hash (with the same information as remote revisions, but an extra "local" string in the input to the hash). This is to avoid conflicts where a remote might gain a package with the same hash as a locally-built package; in that case, the two would conflict. With this patch, the two revisions of the same package should be able to coexist.

Syncing from and to remote stores can always use the remote-specific hashes. Once we have assigned a revision to the package (or reused an existing one), we know which hash (remote or local) to assign and use for the build.

This commit also extracts the hash calculation into a separate function for cleanliness.

The same hashes are still generated for the same specs, so existing tarballs on the remote store are used.

A bit of reorganisation was necessary to move the hash calculation into the main build loop, as we need to know our dependencies' hashes at the time we calculate our own hash, but to know their hashes, they must have first been assigned a revision number (so that we know whether to use the local or remote hash).

I've tested this by building O2 and AliPhysics on slc7; aliBuild successfully reuses existing packages from the store, and builds some locally with localN revisions. I haven't run into a situation where a remote store package was built with the same config as an already-existing local package, but the separate local/remote hashes should solve that. I think the best way to test that is probably to deploy this change on a CI builder for a little while and see how it goes.

Supersedes #666; this PR is split into three commits (see also their descriptions). Each commit passes the unit tests individually; together, they are equivalent to #666 modulo some very minor refactoring of `storeHashes`.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -366,6 +366,51 @@ def createDistLinks(spec, specs, args, repoType, requiresType):
                  t=target)
     execute(cmd)
 
+
+def storeHashes(package, specs, isDevelPkg, considerRelocation):
+  """Calculate various hashes for package, and store them in specs[package]."""
+  spec = specs[package]
+  h = Hasher()
+  dh = Hasher()
+
+  for x in ["recipe", "version", "package", "commit_hash",
+            "env", "append_path", "prepend_path"]:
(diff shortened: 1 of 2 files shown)


ktf:
Thank you for your refactoring effort.

ktf:
Let's try this with one of the failing builders.

PR #668, opened 2021-03-12
Update existing git repos if we can't checkout the tag

TimoWilken:
This fetches the tags of existing git repos if desired tag can't be checked out, then retries checking out the desired tag. This only results in an extra network request per repo (which might mean an extra password prompt) if the tag doesn't already exist, which happens if the tag was made after the repo in `SOURCES` was first cloned.

This doesn't catch cases where a tag moves, or where `$GIT_TAG` is a branch, not a
tag.

### DIFF
--- a/alibuild_helpers/build_template.sh
+++ b/alibuild_helpers/build_template.sh
@@ -92,7 +92,11 @@ if [[ ! "$SOURCE0" == '' && "${SOURCE0:0:1}" != "/" ]]; then
   else
     # Folder is already present, but check that it is the right tag
     cd $SOURCEDIR
-    git checkout "${GIT_TAG}"
+    if ! git checkout "$GIT_TAG"; then
+      # If we can't find the tag, it might be new. Fetch tags and try again.
+      git fetch -f "$SOURCE0" "refs/tags/$GIT_TAG:refs/tags/$GIT_TAG"
+      git checkout "$GIT_TAG"
+    fi
   fi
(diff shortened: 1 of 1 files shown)


alibuild_helpers/build_template.sh
```diff
@@ -92,7 +92,11 @@ if [[ ! "$SOURCE0" == '' && "${SOURCE0:0:1}" != "/" ]]; then
   else
     # Folder is already present, but check that it is the right tag
     cd $SOURCEDIR
-    git checkout "${GIT_TAG}"
+    if ! git checkout "$GIT_TAG"; then
+      # If we can't find the tag, it might be new. Fetch tags and try again.
+      git fetch -f "$SOURCE0" --tags
```
ktf, line 97: Why not fetching just that specific tag?

ktf: ping

TimoWilken: We could fetch just that tag, I don't really mind either way.

TimoWilken: Done.

TimoWilken: Testing the new command, it seems that it fetches all new tags anyway:
```
$ git fetch -f upstream "refs/tags/v1.7.3:refs/tags/v1.7.3" 
From https://github.com/alisw/alibuild
 * [new tag]         v1.7.4.rc2 -> v1.7.4.rc2
```

PR #670, opened 2021-04-06
Use standard name for upstream remotes.

vkucera:


ktf:
I agree on the change, however @mpuccio should comment how much work that is on adapting the tutorials.

mpuccio:
The git part of the tutorial is taught together with LHCb, so it is more abstract. We should update though the [ALICE page on the git workflow ](http://alisw.github.io/git-tutorial/) that is actually not in the tutorial (we should try maybe to make everything homogenous). Btw, while I enjoy teaching at the tutorials, the burden of organising them is on the shoulders of others :)

vkucera:
> The git part of the tutorial is taught together with LHCb, so it is more abstract. We should update though the [ALICE page on the git workflow ](http://alisw.github.io/git-tutorial/) that is actually not in the tutorial (we should try maybe to make everything homogenous). Btw, while I enjoy teaching at the tutorials, the burden of organising them is on the shoulders of others :)

I can propagate the changes there if somebody can point me to the GitHub repos.

ktf:
https://github.com/alisw/alisw.github.io/blob/master/git-tutorial.markdown

PR #671, opened 2021-04-09
Fix aliBuild not detecting that a dev package needs rebuilding

TimoWilken:
For some reason, this was broken recently, such that after changing a development package's sources, it would need two aliBuild invocations to actually be rebuilt. This patch should fix this (and testing with AliPhysics, it does).

I'm slightly worried this might introduce unnecessary rebuilds, but in my testing I haven't come across any (yet?).

I'm still not sure why this bug has suddenly appeared as this particular code hasn't been touched in a while. Perhaps it has to do with the recent reshuffling of the hashing logic.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -964,14 +964,9 @@ def doBuild(args, parser):
                                                spec["version"],
                                                spec["revision"])
     fileHash = readHashFile(hashFile)
-    if fileHash != spec["hash"]:
-      if fileHash != "0":
-        debug("Mismatch between local area (%s) and the one which I should build (%s). Redoing.",
-              fileHash, spec["hash"])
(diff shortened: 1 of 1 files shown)


ktf:
Can you illustrate a case you think it might rebuild unnecessarily?

TimoWilken:
> Can you illustrate a case you think it might rebuild unnecessarily?

Not really, to be honest. I'm just a bit unsure because this particular logic was the same before the recent changes with local revisions. I can't think of a case where this is too aggressive, but I may just have missed something.

pzhristov:
I tested it with changes from github, with local change and without changes: it all three cases it behaves correctly.

pzhristov (approved):

ktf:
Ok. Merging for now since master is currently broken. The regression in the coverage is real, but just because we have now less lines of code.

PR #672, opened 2021-04-12
Improve ability to filter requires

ktf:
This extends the ability to filter requires, currently possible
only on a per architecture basis, to defaults and a newly introduced
"target".
The old syntax:

  requires:
    zlib: osx_x86-64   # Only on macOS

is still valid and equivalent to:

  requires:
    zlib: arch=osx_x86-64   # Only on macOS

One can now also do:

  requires:
    zlib: defaults=o2       # Only when compiling O2

or:

  requires:
    zlib: target=analysis        # Only when invoking `alibuild build O2:analysis`

Multiple filters can now be specified via comma separation,
and will be logically ANDed together.

### DIFF
--- a/alibuild_helpers/utilities.py
+++ b/alibuild_helpers/utilities.py
@@ -168,10 +168,28 @@ def getVersion():
     err, version = getstatusoutput(cmd)
     return version if not err else "Unknown version."
 
-def filterByArchitecture(arch, requires):
+def filterRequires(arch, default, target, requires):
   for r in requires:
-    require, matcher = ":" in r and r.split(":", 1) or (r, ".*")
-    if re.match(matcher, arch):
+    if ":" not in r:
+      yield r
+      continue
+    require, matchers = r.split(":", 1)
+    if not "=" in matchers:
+      matchers = "arch=" + matchers
+    matchesAll = True
+    for matcher in matchers.split(","):
+      if not "=" in matcher:
+        raise "Bad require filter definition"
+      field, matcher = matcher.split("=")
+      if field not in ["arch", "defaults", "target"]:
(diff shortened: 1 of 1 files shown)


alibuild_helpers/utilities.py
```diff
@@ -310,7 +328,12 @@ def getPackageList(packages, specs, configDir, preferSystem, noSystem,
...
   while packages:
-    p = packages.pop(0)
+    fullPkg = packages.pop(0)
+    target = all
+    if ":" in fullPkg:
+      p, target = fullPkg.split(":")
+    else:
+      p = fullPkg
     if p in specs:
       continue
```
TimoWilken, line 338: Should we really `continue` here? What if we need to build packages with conflicting targets?

For example, what should happen with an invocation such as `aliBuild build O2:analysis arrow:some-other-target`, where `O2:analysis` depends on `arrow`?

Should we propagate targets to dependencies, and build them twice if necessary? How do we keep them apart? (A lot of code currently assumes that if `spec1["package"] == spec2["package"]`, then `spec1 == spec2` -- which this would break.)

TimoWilken (commented):
Looks like a good idea to me in general, but I have a few questions:

- How do we handle conflicting targets?
- Should we propagate targets to dependencies?
- Will we need different build scripts for each package depending on what target we're building? Or could we e.g. make an `ALIBUILD_TARGET` variable available to the main build script? Or is this unnecessary?
- Should `spec["target"]` go into the hash calculation? This might depend on how we handle dependencies' targets.

ktf:
Closing. Will rethink about this at some point when I am less busy...

PR #673, opened 2021-04-13
Revert "Use standard name for upstream remotes."

ktf:
Reverts alisw/alibuild#670


ktf:
Reverting this to create the alibuild tag without waiting for the discussion.

PR #674, opened 2021-04-13
Extract sync helpers into separate submodule

TimoWilken:
These were previously in build.py, but don't depend on anything else there. To simplify build.py, extract sync code into sync.py and just import the required classes from build.py.

The code in sync.py is unchanged from the code in build.py (it was only moved over), modulo a few changes (`join` -> `os.path.join` and `basename` -> `os.path.basename`) to simplify imports in sync.py.

Unit tests are passing locally and building zlib with the patched aliBuild works fine.


ktf (approved):
Good to me if it passes tests.

PR #675, opened 2021-04-13
Use a manifest file to speed up symlink downloads

TimoWilken:
With this commit, aliBuild looks for a `<package>.manifest` file listing symlink names and targets, in order to avoid having to download many small files.

The manifest file is a simple text file, so symlink names and targets must not contain newlines, and symlink names must not contain tab characters.

This change is backwards-compatible: the many individual pseudo-symlinks can stay; they will be simply ignored (and not downloaded) if they appear in the manifest. Old aliBuild versions will ignore the manifest and can continue to use the individual pseudo-symlink files.

Needs testing.

### DIFF
--- a/alibuild_helpers/sync.py
+++ b/alibuild_helpers/sync.py
@@ -97,9 +97,8 @@ class HttpRemoteSync:
               # No need to retry any further
               return None
             resp.raise_for_status()
-            resp = resp.text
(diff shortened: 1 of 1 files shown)


ktf (approved):
As discussed this is indeed the way to go. Ok to merge once verified it does not have hidden failure modes.

TimoWilken:
OK, thanks @ktf.

PR #676, opened 2021-04-16
Fix directory listing using S3 HTTPS API

TimoWilken:
Previously, trailing slashes would be stripped from paths. This led to unintended symlinks being created under `TARS`, e.g. when trying to fetch `TARS/slc7_x86-64/O2/`, `TARS/slc7_x86-64/O2` would be used as the prefix instead, leading to O2Suite symlinks being included in the result as well.

We do want to strip leading slashes from the prefix, so use `str.lstrip` instead.


ktf:
Do we need 1.8.1?

TimoWilken:
Not right away. This is really just a slight annoyance at the moment and leads to a few too many symlinks being downloaded individually with the upcoming symlink caching system, but it doesn't cause any other wrong behaviour in aliBuild. I propose we include this in the tag with the symlink caching on Monday.

PR #677, opened 2021-04-23
Create init.sh files in advance.

ktf:
This will create the init files before the build, rather
than after, so that people can use them when a build fails,
to reproduce the issue.


ktf:
Does not really work. Leaving this for better times.

PR #678, opened 2021-04-27
Provide REQUESTED_VERSION to prefer_system_check

ktf:


ktf:
I agree it is probably useful regardless of what we end up doing with GCC.

PR #682, opened 2021-05-03
Allow running tests on unknown architectures

TimoWilken:
- add a `--force-unknown-architecture` to each command that needs it for tests
- update distro versions in the help message that is shown when an unknown architecture is detected
- add recent Debian versions to the Debian -> Ubuntu conversion table
- fix warnings thrown from `alibuild_helpers.utilities.doDetectArch` due to unclosed files when running `uname -m`
- handle the case where `platform.processor() == ""` but `platform.machine()` is fine -- in that case, we don't need to call `uname -m`
- clean up `args` tests a little, testing invalid architectures as well
- clean up tests' imports, so we don't need `mock` installed when using python3

@ktf, with this PR, I can actually run alibuild's tests on my machine instead of making a docker container each time. I cleaned up a few things on the way. There are lots of small changes in here, would it be better if I split it up or is it OK to review like this?

### DIFF
--- a/alibuild_helpers/args.py
+++ b/alibuild_helpers/args.py
@@ -180,29 +180,27 @@ def doParseArgs(star):
   args = finaliseArgs(parser.parse_args(), parser, star)
   return (args, parser)
 
-VALID_ARCHS_RE = ["slc[5-9]+_(x86-64|ppc64)",
-                  "(ubuntu|ubt|osx|fedora)[0-9]*_x86-64",
-                 ]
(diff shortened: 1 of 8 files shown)


ktf:
Ok, but how do we make sure that running the tests on a known architecture does not have a regression?

TimoWilken:
What kind of regression do you mean? I've only changed the tests for argument parsing. CI here is on Ubuntu, which is a known arch, so any other tests run exactly as before.

PR #686, opened 2021-05-04
Autodetect slc6 correctly

TimoWilken:
On slc6, `distribution` is `CentOS` initially, so we need to make it lowercase to match.


ktf (changes requested):
Please check why the tests are failing.

TimoWilken:
There is a step in the tests that builds the jekyll docs. That sometimes fails randomly because some external resources on SourceForge return 503 occasionally. I've restarted the tests.

PR #689, opened 2021-05-11
Additional fix for Mac M1

pzhristov:


alibuild_helpers/utilities.py
```diff
@@ -132,7 +132,11 @@ def format(s, **kwds):
 
 def doDetectArch(hasOsRelease, osReleaseLines, platformTuple, platformSystem, platformProcessor):
   if platformSystem == "Darwin":
-    return "osx_x86-64"
+    systname,nodename,release,version,machine = os.uname()
```
TimoWilken, line 135: Hm, this seems to make the unit tests fail. Would it be possible to prefer `platformProcessor` over what `os.uname()` returns? i.e.
```python
machine = platformProcessor or os.uname()[4]
if machine == ...
```

TimoWilken (changes requested):
Looks good to me in general.

pzhristov:
Something is wrong, now it returns osx_arm64 for Intel Mac. I am closing this PR and will provide a working fix.

PR #693, opened 2021-05-18
Fix python pip installation for python2

ktf:


ktf:
Apparently the new syntax does not work on python2...

PR #699, opened 2021-06-22
Use remote store by default on Ubuntu 20.04

TimoWilken:
We build O2 nightly, as for slc7.

Tested in a fresh `alisw/ubuntu2004-builder` container, with aliBuild installed using `pip3`.

### DIFF
--- a/alibuild_helpers/args.py
+++ b/alibuild_helpers/args.py
@@ -228,7 +228,7 @@ def finaliseArgs(args, parser, star):
     args.configDir = format(args.configDir, prefix="")
 
     # On selected platforms, caching is active by default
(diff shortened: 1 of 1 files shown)


ktf:
Can you test alibuild installed by apt?

TimoWilken:
Building and installing the `.deb` works too -- that aliBuild also uses remote packages.

ktf:
we should probably release a new alibuild, then. Anything else that we need?

TimoWilken:
I'll make a new release. Do we need an entry in debian/changelog?

ktf:
i will take care of that. it's actually not completely straightforward...

PR #701, opened 2021-07-01
Override Docker image entrypoint

TimoWilken:
If the image has an entrypoint, it will be prepended to our command. We don't want that, so explicitly override it to the empty string.

Tested and working in the fullCI checkers.


TimoWilken:
Force-pushed to update the unit test to match. Fine to merge now.

PR #703, opened 2021-07-05
Store version in alibuild_helpers instead of setup.py

TimoWilken:
Currently, recent Python 3 versions will throw a `ResourceWarning` from `pkg_resources` when getting the installed version number in unit tests. This patch fixes that.

Storing the version number internally also means we don't need a fallback in case alibuild isn't installed with pip.

The version number is still only stored once. `setup.py` simply gets it from the same place as the `aliBuild` script.


ktf:
Do you understand why the reduction in coverage?

TimoWilken:
Not really, but the difference is tiny (-0.12%). Maybe removing `getVersion` and its associated test changed the proportions a little.

PR #704, opened 2021-07-05
Show s3cmd debug output for uploads as well

TimoWilken:
For downloads, debug output is already shown.

This should make it easier to debug problems with S3 uploads.


TimoWilken:
This should help debug what's going wrong with the DailyO2-slc7-py3 Jenkins builds. They claim they're uploading tarballs to S3, but none actually appear. I can't reproduce this locally, so hopefully this patch will help debug it.

If it's OK, I'll make a new release once this is merged.

ktf:
Will this trigger a bunch of debug messages for everyone?

TimoWilken:
No, only for people using an `s3://...:rw` remote (not people using `https://...`). That's only people who have S3 credentials and use them to upload packages, which should be pretty rare.

PR #710, opened 2021-07-23
Speed up PR checker

TimoWilken:
- only install as much as we need for each step
- do fast steps first
- inline scripts/cibuild
- run documentation check last, as installing its dependencies is slow
- drop Python 3.5 (slc7-builder has 3.6.8), and add 3.9


ktf (changes requested):
The old script was also taking care of uploading the release. What happened to it?

TimoWilken:
The stuff at the end of `cibuild` depends on a bunch of Travis-specific env vars, so it doesn't run in GitHub actions. And that publishing is done by [`release.yml`](https://github.com/alisw/alibuild/blob/master/.github/workflows/release.yml) now.

TimoWilken:
Sorry for the pings, I'm done debugging now.

TimoWilken:
Conflicts fixed

TimoWilken:
The documentation check keeps failing because https://badge.fury.io/ is down. The other checks are fine.

TimoWilken:
Fixed now!

PR #714, opened 2021-09-29
Fix upload of dist links using s3:// remote

TimoWilken:
This fixes a mistake where aliBuild would upload dist links with their full path as the key, which also breaks RPM publishing.

### DIFF
--- a/alibuild_helpers/sync.py
+++ b/alibuild_helpers/sync.py
@@ -334,7 +334,8 @@ class S3RemoteSync:
     if not self.writeStore:
       return
     execute("""\
-    find '{w}/{t}' -type l | while read -r x; do
+    cd {w} || exit 1
+    find {t} -type l | while read -r x; do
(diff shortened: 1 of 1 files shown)


alibuild_helpers/sync.py
```diff
@@ -334,7 +334,8 @@ def syncDistLinksToRemote(self, link_dir):
     if not self.writeStore:
       return
     execute("""\
-    find '{w}/{t}' -type l | while read -r x; do
+    cd {w} || exit 1
```
ktf, line 337: I thought execute runs with `-e`?

ktf (commented):
Thanks. Now you see why I always suggest to have the smallest possible set of changes... ;-)

alibuild_helpers/sync.py
TimoWilken: It may well do, I didn't check.

Issue #721, opened 2021-11-03
Add option for disabling stores


adriansev:
Hi! Would be really useful if there would be an option for disabling the stores.
There are many instances when one would like to have a full recompilation even on supported official distros.

ktf:
Merged in master. Closing this. Please reopen if you still have troubles.

PR #722, opened 2021-11-03
Allow disabling the remote store completely

TimoWilken:
This patch gives users an option to disable the default remote store.


TimoWilken:
Fixes #721. Not sure what broke the analytics unit tests since they were working before, but https://github.com/alisw/alibuild/pull/722/commits/3d4e07d07a63aabdeb6a1e48fcf0473e530cde75 fixes them.

PR #723, opened 2021-11-03
Improve command-line argument documentation

TimoWilken:
Lots of command-line options had no documentation at all when running `aliBuild <command> --help`. This PR fixes that and makes the existing help text more consistent across commands.

I changed a few small parts of argument parsing:

- I removed the ability to specify a `%(prefix)s` in `--config-dir` for `aliBuild build` -- it was always substituted with the empty string.
  - `aliBuild init` still has this ability, and it substitutes the argument to its `--devel-prefix` option for `%(prefix)s`. Is this useful or can it be removed as well? It seems like `aliBuild build` may have done this in the past as well, but not any more.
  - The meanings of `-z`/`--devel-prefix` in `aliBuild build` and `aliBuild init` have apparently diverged significantly. See above -- is the option still needed for `aliBuild init`?
- `aliBuild clean` now accepts any architecture, even if unsupported -- my reasoning is that if we're cleaning up for a specific architecture, we previously built something for that architecture, so the user wants to use it. I'd say having to use `--force-unknown-architecture` is more of a hassle in this case.

I also updated the synopsis of `aliBuild build` in the user documentation, as it was out of date.

### DIFF
--- a/alibuild_helpers/doctor.py
+++ b/alibuild_helpers/doctor.py
@@ -8,10 +8,9 @@ except ImportError:
 import logging
 from alibuild_helpers.log import debug, error, banner, info, success, warning
 from alibuild_helpers.log import logger
-from alibuild_helpers.utilities import getPackageList, format, detectArch, parseDefaults, readDefaults, validateDefaults
+from alibuild_helpers.utilities import getPackageList, parseDefaults, readDefaults, validateDefaults
 from alibuild_helpers.utilities import dockerStatusOutput
 from alibuild_helpers.cmd import getStatusOutputBash
-import subprocess
 
 def prunePaths(workDir):
   for x in ["PATH", "LD_LIBRARY_PATH", "DYLD_LIBRARY_PATH"]:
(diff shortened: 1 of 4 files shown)


alibuild_helpers/doctor.py
```diff
@@ -72,31 +71,6 @@ def systemInfo():
     if not err:
       debug("%s:\n%s", f, out)
 
-def doctorArgParser(parser):
```
ktf, line 75: This change is not related, is it?

TimoWilken: I moved that with the other code setting up args, so it's now at https://github.com/alisw/alibuild/pull/723/files/d3ba69b04105bea47cb62aba36e7eb53e344bb80#diff-0050151db7b4b98f46fb311611511578bee2f2a821a07665520e0e72e1fadb32L157 in alibuild_helpers/args.py (click "show diff"). This way, all the argparse code is together.

tests/test_args.py
```diff
@@ -76,7 +75,7 @@ class FakeExit(Exception):
   # With ALIBUILD_WORK_DIR and ALIBUILD_CHDIR set
   (("sw2", ".")    , "build --force-unknown-architecture zlib"                         , [("action", "build"), ("workDir", "sw2"), ("referenceSources", "sw2/MIRROR"), ("chdir", ".")]),
   (("sw3", "mydir"), "init"                                                            , [("action", "init"), ("workDir", "sw3"), ("referenceSources", "sw3/MIRROR"), ("chdir", "mydir")]),
-  (("sw", ".")     , "clean --force-unknown-architecture --chdir mydir2 --work-dir sw4", [("action", "clean"), ("workDir", "sw4"), ("referenceSources", "sw4/MIRROR"), ("chdir", "mydir2")]),
```
ktf, line 79: Why was this removed?

TimoWilken: See the PR description:

> aliBuild clean now accepts any architecture, even if unsupported -- my reasoning is that if we're cleaning up for a specific architecture, we previously built something for that architecture, so the user wants to use it. I'd say having to use `--force-unknown-architecture` is more of a hassle in this case.

PR #727, opened 2021-11-10
Fix relocation of libs with multiple LC_RPATH entries

TimoWilken:
Some libraries built on OSX can have LC_RPATH entries for multiple architectures. In that case, they'll have the same rpath multiple times, but a single `install_name_tool` call will change all of them, and subsequent calls will fail.

Instead, only consider each unique rpath only once.


ktf:
Nice catch... @pzhristov this probably solves the issue with GEANT4. @TimoWilken could you do a release including it?

TimoWilken:
@ktf Done!

pzhristov:
@ktf Unfortunately this fix did not solve the issue on Monterey.

PR #728, opened 2021-11-25
Use only single Docker container for running system checks

TimoWilken:
Setting up and cleaning up a Docker container for each check takes annoyingly long, so just spawn a single container and run commands inside it individually.

Tested using `aliDoctor O2 --defaults o2 --docker` on slc7 with python2 and python3.

### DIFF
--- a/alibuild_helpers/analytics.py
+++ b/alibuild_helpers/analytics.py
@@ -1,12 +1,11 @@
 #!/usr/bin/env python
 import os, subprocess, sys
-try:
-  from commands import getstatusoutput
-except ImportError:
-  from subprocess import getstatusoutput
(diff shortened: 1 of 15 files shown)


TimoWilken:
Sorry for the spam. As this turned into a larger reshuffling, I separated each step out into an individual commit. Each commit passes unit tests locally, and I've tested the whole package with `aliDoctor --docker`, `aliBuild deps --docker` and `aliBuild build --docker`.

ktf:
Looks good and it might indeed solve some problems with docker we are also seeing on @costing machines. Is  it good to go or do you want to do more tests? Could you give this top priority? Thanks.

TimoWilken:
These commits were included in #729.

PR #729, opened 2021-11-26
Use host network for Docker builds, and no network for system checks

TimoWilken:
System checks shouldn't need to use the network, so disable it in Docker builds for a speed-up.

For regular builds, we can just use the host's network stack as we don't need any particular network features from Docker, and this way it's faster to set up and tear down containers.

This PR depends on #728, as it includes the commits from that PR. Once #728 is merged, I'll rebase and then this PR can be merged.

### DIFF
--- a/alibuild_helpers/analytics.py
+++ b/alibuild_helpers/analytics.py
@@ -1,12 +1,11 @@
 #!/usr/bin/env python
 import os, subprocess, sys
-try:
-  from commands import getstatusoutput
-except ImportError:
-  from subprocess import getstatusoutput
-from alibuild_helpers.log import debug, banner
 from os.path import exists, expanduser
 from os import unlink
+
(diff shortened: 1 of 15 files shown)


ktf:
Are you  sure that "no network" does not imply creating a dummy one with no access to the outside?

TimoWilken:
`man docker-run` just says that `--network=none` means "No networking in the container".

Running `docker run --rm --network=none alisw/slc7-builder ip a` only shows a loopback interface, so presumably it's disabling most real network functionality completely.

ktf:
Which is what worries me. I assume that on the other hand `--network host` will actually be  the less intrusive way for it to behave. Moreover, we avoid having different behaviour between bare metal and in docker builds.

TimoWilken:
Fair enough, changed.

PR #730, opened 2021-12-01
Cleanup INSTALLROOT

TimoWilken:
As far as I can tell, this directory is only used for temporary storage between building and the proper installation directory, but it's not cleaned up when the build fails.

### DIFF
--- a/alibuild_helpers/clean.py
+++ b/alibuild_helpers/clean.py
@@ -21,10 +21,11 @@ def decideClean(workDir, architecture, aggressiveCleanup):
   """
   symlinksBuild = [os.readlink(x) for x in glob.glob("%s/BUILD/*-latest*" % workDir)]
   # $WORK_DIR/TMP should always be cleaned up. This does not happen only
-  # in the case we run out of space while unpacking.
+  # in the case we run out of space while unpacking. INSTALLROOT is similar,
+  # though it is not cleaned up automatically in case of build errors.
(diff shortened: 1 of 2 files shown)


ktf:
The idea was that in case of build failures we want to have it around and eventually clean it up by hand via aliBuild clean. Maybe we should change the build script to invoke aliBuild clean more often?

TimoWilken:
> The idea was that in case of build failures we want to have it around and eventually clean it up by hand via aliBuild clean.

The first part happens currently -- `INSTALLROOT` is kept on build failures -- but the second part does not. `aliBuild clean` without this patch does not clean up `INSTALLROOT`.

> Maybe we should change the build script to invoke `aliBuild clean` more often?

The script already calls `aliBuild clean` before and after every build. The problem is that currently, that doesn't touch `INSTALLROOT` at all.

ktf:
Ah, ok, I missed this was for `aliBuild clean`. This does not interfere with the development packages does it?

ktf (commented):
Ah, ok, I missed this was for aliBuild clean. This does not interfere with the development packages does it?

TimoWilken:
As far as I know, development packages are installed directly into `sw/$arch/`; they don't go through `INSTALLROOT` at all: https://github.com/alisw/alibuild/blob/1ea51fbaf8c31934aab0993465fee38a6638387b/alibuild_helpers/build_template.sh#L50-L57

PR #731, opened 2021-12-02
Fix encoding errors from getstatusoutput

TimoWilken:
This happened on python3.6, as there, the system encoding (which for `LC_ALL=C` is ASCII) was used to decode command output. However, Vc has some branch names with UTF-8 non-breaking spaces in them. Reimplementing the function allows specifying an explicit encoding, which fixes this.

Tested on python3.6 using `LC_ALL=C`.


ktf:
Thanks. Please make a new release.

PR #733, opened 2021-12-03
Make getstatusoutput behaviour match library function

TimoWilken:
`subprocess.getstatusoutput` strips a single trailing newline, if one exists. Some calling code relies on this.


ktf:
Please tag again 1.9.6 is broken.

PR #735, opened 2021-12-18
Avoid openssl interference with git

ktf:
For some reason, the dependencies environment is actually present here, potentially making the system git misbehave on cs8. In order to avoid this, we run git in a completely pristine  environment, so that interference  is avoided (at the cost of not working with custom git installations).


ktf:
@TimoWilken do you understand why this is actually needed in order to compile OpenSSL 1.1.1m?

PR #737, opened 2022-01-19
Fix otool invocation for multi-arch binaries

TimoWilken:
For "universal files", i.e. binaries with multiple architectures, otool displays all architectures' values, which confuses the script. Make it show only the arch we're building.

Tested and working on osx_arm64.


ktf:
Nice, thanks. We should probably do a release with this (and most importantly update homebrew). Can you take care of it?

TimoWilken:
OK, done!

PR #739, opened 2022-01-26
Rebuild later packages if dev package changes

TimoWilken:
If a development package has committed or uncommitted changes, any packages depending on it should be rebuilt with the new code. This patch implements this.

This is done by incorporating the `devel_hash` of every dependency (if present) in a package's overall hash, so that if a dependency changes, we rebuild packages that depend on it. A dev package's hash itself doesn't change, however, so we keep rebuilding it in the same build area.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -170,8 +170,9 @@ def storeHashes(package, specs, isDevelPkg, considerRelocation):
   dh = Hasher()
   for dep in spec.get("requires", []):
     # At this point, our dependencies have a single hash, local or remote.
-    h_all(specs[dep]["hash"])
-    dh(specs[dep]["hash"] + specs[dep].get("devel_hash", ""))
+    hash_and_devel_hash = specs[dep]["hash"] + specs[dep].get("devel_hash", "")
+    h_all(hash_and_devel_hash)
+    dh(hash_and_devel_hash)
 
   if isDevelPkg and "incremental_recipe" in spec:
     h_all(spec["incremental_recipe"])


TimoWilken:
@ktf is this approach OK? Can you see problems with it? It would invalidate the hashes of e.g. O2Physics from the existing nightly tags, but not O2 or (hopefully) AliPhysics, as those don't have dev package dependencies (though I don't know in the latter case).

ktf:
This is for sure the most conservative approach, but in the past there were complaints it  basically makes whatever depends on a low level package to rebuild from scratch, even if the change was just an implementation detail. What I do not understand is: if there is a commit for a devel package, the behaviour should already be the one you try to force here. Is it not the case?

TimoWilken:
> This is for sure the most conservative approach, but in the past there were complaints it basically makes whatever depends on a low level package to rebuild from scratch, even if the change was just an implementation detail.

Let's say you're changing autotools, which is a dev package. You then want to rebuild autotools, but not ROOT, O2, etc etc, because you're only changing an implementation detail in autotools. In that case, why not run `aliBuild build autotools` instead of `aliBuild build O2`? Then only autotools will be built, and because it's a dev package, any existing alienv environment should pick up the newly rebuilt version automatically, right?

> What I do not understand is: if there is a commit for a devel package, the behaviour should already be the one you try to force here. Is it not the case?

No. Currently (without this PR), the hash of a package depends on the `spec["hash"]` of its dependencies. But for development packages, we want the same `spec["hash"]` so that we can rebuild in the same build area as before.

For example, O2's `spec["hash"]` depends on its `spec["commit_hash"]`, which is forced to `"0"` for dev packages. Then, if the O2 repo changes, its `spec["commit_hash"]` will still only be `"0"`, so its hash won't change. Its `spec["devel_hash"]` will change, but we don't currently use that for the overall hash calculation.

`spec["devel_hash"]` is only checked *for the package itself* (i.e. for O2) to see whether it needs to be rebuilt. For dependent packages (e.g. QC), we don't check dependencies' `spec["devel_hash"]`.

ktf:
Ok, it makes sense. Maybe we can introduce this as a command line option as @davidrohr was suggesting?

ktf (approved):

ktf:
Ok, I guess that while this will make some previous quick builds become slow, it's the lesser of the possible evils we can think of... The quickest way to juggle multiple dev packages will remain simply cd-ing in the folder(s) and typing "ninja" by hand. I guess we write that somewhere. Merging.

PR #741, opened 2022-02-03
Use a connection pool for HTTP remote stores

TimoWilken:
This way, we only connect to a HTTP remote once, and then keep reusing this connection for each request. This should give a speed-up, especially for the many small requests when listing directories.

@ktf: Is it OK to keep a single connection open for the lifetime of the process, i.e. potentially hours? Or should aliBuild instead open a new connection for syncing each package individually, in which case we'd have to open 2n connections, where n is the number of packages needed, but each connection would only last a few seconds?


ktf:
maybe a first step would be to open a pool per package?

PR #742, opened 2022-02-04
Use GIT_DIR rather than cd

ktf:


adriansev:
@ktf just curious, why is there a check for python 2.x? i do not think that any other ALICE component support or require it... and it restricts a lot the code, and force the usage of different code for the same thing, which is suboptimal ...

ktf:
To be compatible with production releases which were done for slc6.

TimoWilken:
Looks like this breaks the tests because the test `alidist` we use is in the alibuild repo, so it doesn't have a `.git` of its own. We could:

1. require a toplevel `.git` in repos, and factor out `alibuild/tests/testdist` into another repo
2. use `git -C %s rev-parse HEAD` instead, but git on slc6 doesn't have a `-C` option
3. go back to `cd %s >/dev/null 2>/dev/null && git rev-parse HEAD`

ktf:
Ok, 3. and maybe add a comment about it.

PR #744, opened 2022-02-07
Introduce support for plugins

ktf:


ktf:
@TimoWilken could you cut a release?
@davidrohr could you check the behavior of the devel packages is as you requested?

PR #748, opened 2022-02-10
Avoid overriding the disable packages

ktf:


TimoWilken (approved):

TimoWilken:
I didn't realise `list.copy` doesn't exist on python2. Changed to `list(disable)` instead, which works on both py2 and py3.

PR #749, opened 2022-03-21
Fix variable name in storeHashes

TimoWilken:
I must have missed this when extracting this code out into a function. So far it worked because everywhere `hash_data_for_key` is called, it can access `x` from the outer scope, but it should use its argument directly.


ktf:
Does this mean that we will rebuild everything which used the old version?

TimoWilken:
No, it doesn't. `x` and `key` refer to the same value, but `key` is the correct way to refer to it.

PR #750, opened 2022-03-24
Use a wrapper script that sanitises the environment for OpenSSL-sensitive executables

TimoWilken:
On CentOS 8, when we build our own OpenSSL from alidist, then subsequent git and curl calls fail, as they were compiled against the system OpenSSL, which has some special patches applied.

Therefore, install some wrapper scripts during the build that remove `LD_LIBRARY_PATH` and `DYLD_LIBRARY_PATH` from the environment temporarily for those programs.

When we call git from aliBuild itself (rather than from build scripts), also sanitise the environment. For this, I centralised (almost all) git calls to a single function.

This wrapping of calls needs to be done in two places, as we might be building inside a container, but we always run aliBuild's git commands on the host.

### DIFF
--- a/alibuild_helpers/build_template.sh
+++ b/alibuild_helpers/build_template.sh
@@ -12,6 +12,25 @@ set +h
 function hash() { true; }
 export WORK_DIR="${WORK_DIR_OVERRIDE:-%(workDir)s}"
 
+# Insert our own git and curl wrapper scripts into $PATH, patched to use the
+# system OpenSSL, instead of the one we build ourselves.
+tmpdir=$(mktemp -d)
+export PATH="$tmpdir:$PATH"
+trap -- 'rm -rf "$tmpdir"' EXIT  # clean up on exit
+for executable in git curl; do
+  system_executable=$(command -v "$executable")
+  # Create a wrapper script that cleans up the environment, so we don't see the
+  # OpenSSL built by aliBuild.
+  cat << EOF > "$tmpdir/$executable"
+#!/bin/sh
+exec env -u LD_LIBRARY_PATH -u DYLD_LIBRARY_PATH $system_executable "\$@"
+EOF
+  chmod +x "$tmpdir/$executable"
+  if [ "$(command -v "$executable")" != "$tmpdir/$executable" ]; then
+    echo "WARNING: failed to install $executable wrapper script" >&2
+  fi
+done
+
 # From our dependencies
 %(dependencies)s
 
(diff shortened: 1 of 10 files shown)


TimoWilken:
So far, I've tested this by building O2 with o2 defaults on slc8, both inside a container, and also by using aliBuild's `--docker` option to use an slc8 container.

Probably needs a little more testing, e.g. for AliPhysics and other platforms.

alibuild_helpers/build_template.sh
```diff
@@ -12,6 +12,25 @@ set +h
 function hash() { true; }
 export WORK_DIR="${WORK_DIR_OVERRIDE:-%(workDir)s}"
 
+# Insert our own git and curl wrapper scripts into $PATH, patched to use the
+# system OpenSSL, instead of the one we build ourselves.
+tmpdir=$(mktemp -d)
```
ktf, line 17: I would do it somewhere in the $WORKDIR.

alibuild_helpers/git.py
```diff
@@ -1,4 +1,6 @@
+import shlex
```
ktf, line 1: Is this available also in python2.7?

ktf:
Do we really need to have the new `git()` function call? can't we simply make sure the wrapper is in path?

alibuild_helpers/build_template.sh
TimoWilken: You're right, that's much better.

alibuild_helpers/git.py
TimoWilken: No, you're right, I'll change it to use `pipes.quote` on python2, like elsewhere we use the same function.

TimoWilken:
> Do we really need to have the new `git()` function call? can't we simply make sure the wrapper is in path?

As discussed: I prefer the `git()` function, as it makes unit testing simpler -- we can just mock out `git()` and make it return what we want, rather than having to parse `getstatusoutput` calls.

TimoWilken:
I'm preparing a commit to address your comments above -- I'll test it locally a bit and then push.

ktf:
great, thanks.

TimoWilken:
I've been using the changes in the latest commit to test my changes in alisw/alidist#4019. This works for me both using `--docker` and on the host.

It would probably be good to merge this before alisw/alidist#4019, so the CI for the alidist PR can run with the new alibuild, to make sure it works everywhere too.

PR #751, opened 2022-03-28
Rebuild dev packages incrementally when other dev package changes

TimoWilken:
When a development package depends on another dev package, then when the latter changes, we should rebuild the former incrementally instead of fully.

Cc: @davidrohr

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -175,9 +175,15 @@ def storeHashes(package, specs, isDevelPkg, considerRelocation):
 
   dh = Hasher()
   for dep in spec.get("requires", []):
-    # At this point, our dependencies have a single hash, local or remote.
+    # At this point, our dependencies have a single hash, local or remote, in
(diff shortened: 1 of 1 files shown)


ktf:
Can we make a release with this but then try it with one of the builders before we fully deploy it everywhere?

davidrohr:
I agree, I would roll this out slowly. I just did 2 compile tests :)

TimoWilken:
OK, I'll release v1.10.2.rc1 and try it on a few builders first.

TimoWilken:
So, I've tested this change on a mesosdaq builder, where we have both O2 and QC as dev packages. There, O2 is rebuilt only when the upstream repo's dev branch changes, as expected. QC is rebuilt every time, as the same repo is switched from one PR to the next, so the code changes every time.

With that, I'd say this is good to go -- I'll make a "proper" alibuild release (not a prerelease) and send a PR to update the CI builders to the new version.

PR #753, opened 2022-04-06
Fix errors on slc6

TimoWilken:
Don't append an empty argument if we can't use `--filter=blob:none`.


TimoWilken:
I've published v1.11.1.

PR #754, opened 2022-04-12
Add a templating plugin

TimoWilken:
This allows access to the program state just before the build would start, exposed to a user-provided Jinja2 template.

Required for https://github.com/alisw/ali-bot/pull/1153.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 requests
 pyyaml
 distro
+jinja2
 boto3; python_version >= '3.6'
 futures; python_version == '2.7'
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,6 +1,7 @@
 requests
 pyyaml
 distro
+jinja2
```
ktf, line 4: I assume it supports python2, correct?

TimoWilken: Yes, it does. I tried it under python2.

ktf:
If you are happy, please move it to non WIP.

PR #756, opened 2022-04-19
Add command-line option to force rebuilding specific packages

TimoWilken:
This is an alternative to adding `force_rebuild: true` to the alidist recipe, so that alidist doesn't have to be changed.

Example:

```bash
aliBuild build O2Physics --defaults o2 --force-rebuild O2 --force-rebuild O2Physics
```

(You can specify multiple packages to rebuild by giving the option multiple times. This makes it consistent with how `--disable` works.)

@adriansev: could you check whether this does what you need?


adriansev:
at least from description is what i need, but to actually install the alibuild in devel mode to check it out it will take some time .. 
as a side comment, i'm surprised about the comment related to `--disable` i thought that it worked with `--disable package1,package2,..`

TimoWilken:
> at least from description is what i need, but to actually install the alibuild in devel mode to check it out it will take some time ..

OK, thanks @adriansev!

> as a side comment, i'm surprised about the comment related to `--disable` i thought that it worked with `--disable package1,package2,..`

You're right, that does work. I've made it consistent in the latest commit, so you can use `--force-rebuild O2,O2Physics` as well.

adriansev:
so, i will just leave here as a log the problems encountered and at the end i will ping you...
so, installing with : 
`python3 -m pip install git+file:///home/physics-tools/alice/alibuild@test756`
where test756 is the branch with the PR #756 
gave me:
```
Building wheels for collected packages: alibuild
  Building wheel for alibuild (setup.py) ... done
  Created wheel for alibuild: filename=alibuild-LAST_TAG-py3-none-any.whl size=74958 sha256=a9f78247af9dc3c47718fae8b3e182e585ec4e0d49944c9733d8614d8f1ce9d1
  Stored in directory: /home/adrian/tmp/pip-ephem-wheel-cache-hdvovsx0/wheels/d9/cd/43/19c7e89174be81cc63b9f34d94338526fc22dcdf3a024cb13f
  WARNING: Built wheel for alibuild is invalid: Metadata 1.2 mandates PEP 440 version, but 'LAST-TAG' is not
Failed to build alibuild
Installing collected packages: alibuild
  Attempting uninstall: alibuild
    Found existing installation: alibuild 1.11.1
    Uninstalling alibuild-1.11.1:
      Successfully uninstalled alibuild-1.11.1
  Running setup.py install for alibuild ... done
  DEPRECATION: alibuild was installed using the legacy 'setup.py install' method, because a wheel could not be built for it. A possible replacement is to fix the wheel build issue reported above. Discussion can be found at https://github.com/pypa/pip/issues/8368
```

moreover, something is wrong with the shebang adaptation as i got this:
```
head  /home/adrian/.local/bin/aliBuild
#!python
from __future__ import print_function
import os
import sys
```

after modification by hand to a proper line (`#!/usr/bin/env python3`) the building started successfully

adriansev:
@TimoWilken i can confirm that it works as advertised: after the initial build, another build command will tell me that `Packages already built using this version` but with force-rebuild it will actually rebuild the package. Thanks a lot!

alibuild_helpers/args.py
```diff
@@ -89,14 +86,23 @@ def doParseArgs(star):
                             help=("Fetch updates to repositories in MIRRORDIR. Required but nonexistent "
                                   "repositories are always cloned, even if this option is not given."))
 
-  build_parser.add_argument("--no-local", dest="noDevel", metavar="PKGLIST", default="", type=csv_list,
+  build_parser.add_argument("--no-local", dest="noDevel", metavar="PACKAGE", default=[], action="append",
```
ktf, line 89: Does this mean it will not support comma separated lists anymore? I do not think this is acceptable.

ktf (changes requested):

alibuild_helpers/args.py
TimoWilken: No, 2ebab24515ab9087c4c3998e5e5cb94589ca46a7 fixes this so that `--no-devel` and `--force-rebuild` work like `--disable` (in that `--disable A,B` and `--disable A --disable B` are equivalent).

ktf (approved):

PR #759, opened 2022-05-10
Override Docker image entrypoint for DockerRunner too

TimoWilken:
Running commands in Docker images with non-empty entrypoints fails. This fixes that issue.

Cc: @adriansev

### DIFF
--- a/alibuild_helpers/cmd.py
+++ b/alibuild_helpers/cmd.py
@@ -92,7 +92,7 @@ class DockerRunner:
   def __enter__(self):
     if self._docker_image:
       # "sleep inf" pauses forever, until we kill it.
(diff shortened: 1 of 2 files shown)


TimoWilken:
Hi @adriansev, sorry for the delay. I think that's another issue -- the architecture detection is done on the host, not in the image (which should probably be fixed). So you currently have to specify the architecture of the image as well -- e.g. `-a slc7_x86-64 --docker-image alisw/slc7-builder:latest`.

As a side question, why do you use `-v /home:/home -v /home.hdd:/home.hdd`? That shouldn't be necessary as aliBuild should mount the required paths itself. Is it not doing that?

PR #760, opened 2022-05-13
Improve disable by architecture

ktf:


ktf:
@davidrohr @martenole @awegrzyn: this should allow us to say that we do not want to bother about OpenSSL on anything but slc7. Either by stating:

```
architeture: slc7
```

which will simply disable the package on anything which is not slc7, or by saying:

```
system_requirement: slc7
```

which will consider the `prefer_system_check` (or alternatively `system_requirement_check`) as mandatory for the build to continue.

@awegrzyn could you please have a look? In principle this allows us to drop the `system-XYZ` packages as well, but IIRC you depend on those to create the RPMs. Is that actually the case?

ktf:
@TimoWilken could you cross check this makes sense?

davidrohr:
@ktf : thx, I think this makes a lot of sense (haven't tested it myself though).

alibuild_helpers/utilities.py
```diff
@@ -377,6 +377,18 @@ def getPackageList(packages, specs, configDir, preferSystem, noSystem,
...
+    # Architecture is used to determine for which architecture we
+    # should build a given package. By default everything matches
+    # The architecture string should NOT go in the has, because in
+    # any case wether or not a package is build the hash will be different.
+    architectureRE = spec.get("architecture", ".*")
+    try:
+      architectureREMatches = re.match(architectureRE, architecture)
+    except:
+      dieOnError(True, "Malformed entry architecture: %s in %s" % (architecture, spec["package"]))
+    if architectureREMatches == "False":
```
TimoWilken, line 389: Is this correct? Maybe just `if not architectureREMatches` instead?

alibuild_helpers/utilities.py
```diff
@@ -399,10 +411,10 @@ def getPackageList(packages, specs, configDir, preferSystem, noSystem,
       else:
         ownPackages.update([spec["package"]])
 
-    dieOnError(("system_requirement" in spec) and recipe.strip("\n\t "),
-               "System requirements %s cannot have a recipe" % spec["package"])
     if re.match(spec.get("system_requirement", "(?!.*)"), architecture):
-      cmd = spec.get("system_requirement_check", "false")
+      # system_requirement fallsback to the prefer_system_check rule
+      # if the system_requirement_check does not exist.
+      cmd = spec.get("system_requirement_check", spec.get("prefer_system_check", "false"))
```
TimoWilken, line 417: The `prefer_system_check` expects a `REQUESTED_VERSION` variable, should we add that to this check too?

TimoWilken (commented):
This looks useful to me!

PR #764, opened 2022-05-25
Expose defaults as ALIBUILD_DEFAULTS env variable to the recipe

ktf:


ktf:
@sawenzel this should provide what you requested.

PR #771, opened 2022-08-11
alibuild recipe (reference) : requirements

maireiphc:
Adding extra details about the logic I understood for the fields : 

- requires, 
- build_requires,
- runtime_requires

This comes after some mail exchanges with Adrian Sevcenco, Giulio Eulisse + some local tests on one alidist recipe (lhapdf.sh).
To be revised by Giulio to make sure that this is correct.

### DIFF
--- a/docs/reference.markdown
+++ b/docs/reference.markdown
@@ -88,7 +88,7 @@ The following entries are optional in the header:
     software to be built. This can be a tag name, a branch name or a commit
     hash.
   - `env`: dictionary whose key-value pairs are environment variables to be set
-    after the recipe is built. The values are interpreted as the contents of a
+    _after_ the recipe is built. The values are interpreted as the contents of a
     double-quoted shell string, so you can reference other environment variables
     as `$VARIABLE`, which will be substituted each time another recipe is built.
     For example:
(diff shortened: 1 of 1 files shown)


ktf:
Thank you for this. We should probably check the the runtime_requires behave correctly. I do no think we ever used them...

maireiphc:
I confess that I discovered the `runtime_requires` feature by re-reading your text a few hours ago (in the env variables).
I extrapolated the text, I did not test the thing.
I have been so far unaware of such a thing to exist. I indeed never stumbled upon `runtime_requires` in any recipe. At the moment, there is no occurrence of it among our alidist recipes.

docs/reference.markdown
```diff
@@ -129,7 +131,9 @@ The following entries are optional in the header:
       ...
     ```
 
-    The specified dependencies will be built before building the given package.
+    The specified dependencies will be built before building the given package; 
+    the "required" corresponding binaries, libraries and environment variables 
+    will be made available (by loading the dependency module) before the current built starts.
```
maireiphc, line 136: Do you confirm that a `requires` or `build_requires` load the dependency module before the build starts ?
I mean, it is a "casual" loading of module, there is nothing extra or specific (bin, lib, env) due to the fact that we are about to compile a new package ? In other words, loading a module, be it for a coming built-time or run-time, is the same ?
(I was not sure about this).

ktf: not the dependency module, they load the build environment of the dependencies. aliBuild has no idea about modulefiles.

PR #772, opened 2022-08-16
Use o2 defaults by default, if no option is given

TimoWilken:
With this patch, when users do not specify `--defaults` explicitly, `--defaults o2` is assumed. Previously, `--defaults release` was assumed.

### DIFF
--- a/alibuild_helpers/args.py
+++ b/alibuild_helpers/args.py
@@ -67,7 +67,7 @@ def doParseArgs(star):
   build_parser.add_argument("pkgname", metavar="PACKAGE", nargs="+",
                             help="One of the packages in CONFIGDIR. May be specified multiple times.")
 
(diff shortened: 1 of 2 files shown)


ktf:
This is fine, but we should make sure that the Jenkins action to build AliPhysics releases uses "release".

TimoWilken:
That's already explicitly specified in Jenkins, so it should be fine.

TimoWilken:
@ktf is this OK to merge?

ktf:
@TimoWilken @ddobrigk Please, amend the build instructions to take this into account. No need to have people know about defaults...

ddobrigk:
Thanks, @TimoWilken @ktf - I will take this into account in the documentation (and a standard good default is of course great :-) )

PR #775, opened 2022-08-19
Allow users to input git credentials without prompts overwriting each other

### LINKED ISSUE #681: aliBuild does not correctly get my cern gitlabs credentials from the command line for FLUKA
Running `aliBuild --defaults o2 build O2Suite --debug` on the EPN, it prompts me for my cern credentias to get fluka from gitlab.
Apparently it does a parallel checkout for FLUKA and FLUKA_VMC, which breaks the parsing of the credentials from the command line.
I think what happens is that it asks for my username twice in parallel, overwriting each other.
When I enter the username once, it prints to the console that it asks for the password.
But actually, the second username query is still ongoing. I.e. instead of the password, I have to enter the username again.
Doing so, it asks for the password a second time. So now I enter the password, and alibuild continues.
At some point it gets stuck, and this is because the second password querry still seems to run in the background.
So when I enter the password again, it continues.
This is quite confusing behavior, and it will probably prevent many users from building O2 currently.

TimoWilken:
If a git repository cannot be accessed in parallel, try it sequentially while allowing the user to input their credentials. This fixes the issue where multiple git processes running in parallel would all present a username and password prompt, overwriting each other.

Avoiding password prompts entirely should be out of scope for aliBuild, as that requires changing the user's git configuration. Instead, I updated the documentation to specify how to cache passwords or use SSH auth instead.

Fixes #681

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -30,6 +30,7 @@ try:
 except ImportError:
   from pipes import quote  # Python 2.7
 
+import concurrent.futures
 import importlib
 import socket
 import os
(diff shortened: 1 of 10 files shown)


ktf:
As discussed, we should build a pip release with this and try it out in one of the CI checkers before we move the packaged versions to use it.

TimoWilken:
I set one of the slc7 CI checkers to use this earlier today, and I haven't seen a problem with it since.

PR #778, opened 2022-09-27
Fix tarball unpacking

TimoWilken:
These changes are what I've been testing for the daily builds, and they have fixed the tarball unpacking problems.

It seems the problem must've been `pigz`, as that's the only change in the unpacking step, but I'm not sure why this causes problems randomly when unpacking small tarballs.


TimoWilken:
I've just re-run yesterday's daily build of O2PDPSuite on Ubuntu with this patch applied, rebuilding O2 and everything that depends on it.

The build time is unfortunately a little difficult to compare, as I realised after the build that yesterday's daily had an issue fetching a repo, but if I eliminate the part of the rebuild that was doing the same work as the first build, then I get:

| aliBuild version | Time |
| :-- | --: |
| aliBuild v1.12.0 | 3h04m |
| code from this PR | 2h52m |

...so about a 10-minute speedup due to the parallel copy-and-pack operation. For users, I expect the speedup to be better than that, as they won't be producing tarballs at all.

TimoWilken:
@ktf tested and ready to merge in my view.

PR #780, opened 2022-10-21
Added --dev support

mvala:
Signed-off-by: Martin Vala <mvala@saske.sk>

### DIFF
--- a/alienv
+++ b/alienv
@@ -41,13 +41,14 @@ Usage: $0 \\
     ${EM}help${EZ}
       This help screen.
 
-    ${EM}enter${EZ} [${ET}--shellrc${EZ}] MODULE1[,MODULE2...]
+    ${EM}enter${EZ} [${ET}--shellrc${EZ}] [${ET}--dev${EZ}] MODULE1[,MODULE2...]
       Enters a new shell with the given modules loaded.
       Return to the clean environment by exiting the shell with ${ET}exit${EZ}.
       Inside the environment you can use the native ${ET}modulecmd${EZ}.
       By default you enter the same shell type you are in. Override with environment variable ${ET}MODULES_SHELL${EZ}.
       The new shell will not load your shell startup file by default (e.g. ~/.bashrc) to avoid environment conflicts.
       If you want to retain your shell startup files use ${ET}--shellrc${EZ}.
+      If you want to add environment variables for development use ${ET}--dev${EZ}
       Alternatively you can use the ${EM}load${EZ} command in the current shell.
 
     ${EM}setenv${EZ} MODULE1[,MODULE2...] ${ET}-c${EZ} cmdInEnvironment [PARAM1 [PARAM2...]]
@@ -167,6 +168,7 @@ DEFAULT_WORK_DIRS=( "$ALIBUILD_WORK_DIR" "$ALICE_WORK_DIR" "$ALIBUILD_CHDIR/sw" 
 ARGS=()
 COMMAND_IN_ENV=()
 CLEAN_ENV=1
+DEVOPT=0
 VERBOSE=1
 
 while [[ $# -gt 0 ]]; do
@@ -176,6 +178,7 @@ while [[ $# -gt 0 ]]; do
     --no-refresh) NO_REFRESH=1; shift ;;
     -q) unset VERBOSE; shift ;;
     --shellrc) unset CLEAN_ENV; shift ;;
+    --dev) DEVOPT=1; shift ;;
     --help|help) printHelp; exit 0 ;;
     -c) shift; COMMAND_IN_ENV=("$@"); break ;;
     *) ARGS+=("$1"); shift ;;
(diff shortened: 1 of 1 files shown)


TimoWilken:
Hi @mvala! Thanks for this contribution.

What exactly is the usecase for this new `--dev` option?

If it's intended for creating an environment in which the given software can be compiled, then I'm not sure it will work properly, as our modulefiles generally do not pull in packages' build-only dependencies.

mvala:
> Hi @mvala! Thanks for this contribution.
> 
> What exactly is the usecase for this new `--dev` option?
> 
> If it's intended for creating an environment in which the given software can be compiled, then I'm not sure it will work properly, as our modulefiles generally do not pull in packages' build-only dependencies.

It should create/enter same environment as user would be in recipe and it would be able single package without aliBuild. This feature would be more useful when used against packages located in cvmfs. Then user would do
```
alienv enter --dev <package-dep>
```
And then build `package` locally and deps are going to take from cvmfs

alienv
```diff
@@ -253,22 +257,46 @@ case "$ACTION" in
     MODULES=$(normModules "${ARGS[@]}")
     collectModules
     existModules $MODULES
-    eval $($MODULECMD bash add $MODULES 2> >(grep -v "$IGNORE_ERR" >&2))
+    PS1DEV=
+    if [[ $DEVOPT == 1 ]];then
+      PS1DEV=" (dev)"
+      for MODULE in "$MODULES"; do
+        # echo "source $WORK_DIR/$ARCHITECTURE/$MODULE/etc/profile.d/init.sh"
+        eval $($MODULECMD bash add $MODULES 2> >(grep -v "$IGNORE_ERR" >&2))
```
TimoWilken, line 265: I don't think it makes sense to *both* load the modulefile *and* source the `init.sh` script. They should declare the same variables anyway, and if they are incompatible, the conflicts would be a bit of a mess to sort out. Additionally, when `aliBuild` builds a package, it only sources the `init.sh` script, not the modulefile, so if you want to replicate the build environment as `aliBuild` would create it, then only source `init.sh`.

alienv
```diff
@@ -253,22 +257,46 @@ case "$ACTION" in
     MODULES=$(normModules "${ARGS[@]}")
     collectModules
     existModules $MODULES
-    eval $($MODULECMD bash add $MODULES 2> >(grep -v "$IGNORE_ERR" >&2))
+    PS1DEV=
+    if [[ $DEVOPT == 1 ]];then
+      PS1DEV=" (dev)"
+      for MODULE in "$MODULES"; do
+        # echo "source $WORK_DIR/$ARCHITECTURE/$MODULE/etc/profile.d/init.sh"
```
TimoWilken, line 264: This comment doesn't add anything; please remove it.

alienv
```diff
@@ -253,22 +257,46 @@ case "$ACTION" in
...
+    fi
     printf "Use ${EM}exit${ET} to exit this environment.${EZ}\n" >&2
     exec ${CLEAN_ENV:+env $SHELL_NORC_ENV} $MODULES_SHELL ${CLEAN_ENV:+$SHELL_NORC_PARAM} -i
   ;;
   printenv|load|unload)
+    if [[ $DEVOPT == 1 ]];then
+      echo "*** Please run following commands to load dev environment for '$MODULES' ***"
+      echo
+      echo "export WORK_DIR=$WORK_DIR"
+      for MODULE in "$MODULES"; do
```
TimoWilken, line 294: See above -- do not quote `$MODULES` here. If there are multiple modules, they'll be space-separated in this variable, and if you quote it, then bash won't iterate over the modules.

alienv
```diff
@@ -279,6 +307,7 @@ case "$ACTION" in
     exit 0
   ;;
   setenv)
+    [[ $DEVOPT == 1 ]] && echo "Warning: Development mode for '$ACTION' is not supported. Please remove '--dev'"
```
TimoWilken, line 310: Please make sure user-facing messages go to stderr, not stdout.

TimoWilken (changes requested):
Hi @mvala, thanks. That would indeed be useful, and I see that you source the generated `init.sh` files instead of relying on modulefiles for this, which is good. I have some comments below -- if you'd prefer that I clean the script up and merge it, let me know.

alienv
mvala: I agree with this and i will remove it

mvala:
@TimoWilken I did script cleanup. Hope it is fine now

alienv
```diff
@@ -253,22 +257,44 @@ case "$ACTION" in
...
+    if [[ $DEVOPT == 1 ]];then
+      echo "*** Please run following commands to load dev environment for '$MODULES' ***" >&2
+      echo "" >&2
+      echo "export WORK_DIR=$WORK_DIR" >&2
+      for MODULE in $MODULES; do
+        echo "source $WORK_DIR/$ARCHITECTURE/$MODULE/etc/profile.d/init.sh" >&2
+      done
+      echo "" >&2
+      exit 0
+    fi
```
TimoWilken, line 297: This should definitely go after the verification and normalisation steps -- i.e. after the `MODULES=$(normModules ...)` and `collectModules` lines, so that your `$MODULES` variable has the correct format.

I see you've also changed the script so that *all* the lines go to stderr -- this again means that the following won't do what you might expect:

```bash
eval "$(alienv --dev load MyPackage/v1)"
```

...because it'll print the environment it gets from the modulefiles on stdout, which bash will eval, but not the development environment.

Ideally, this should probably be handled a bit like the `enter` case, where you *either* output `source .../init.sh` lines *or* call `modulecmd`, though you'd have to generate the correct syntax for each possible `$MODULES_SHELL`, keeping in mind that `source` is a bash- and zsh-specific extension (the portable POSIX syntax is just a plain dot, `.`).

If that's all a bit too complicated, I'd recommend just outputting a warning like you do for `setenv`. When the `$ACTION` is `unload`, you'd definitely need this warning, as I'm not sure resetting the environment is possible easily.

TimoWilken (changes requested):
Hi @mvala, thank you very much. I just have a few more concerns about the `load` case (see below) -- but everything else looks good to me now!

TimoWilken (approved):
OK, that works as a compromise. Thanks!

PR #781, opened 2022-10-25
Always rebuild if there are untracked changes

TimoWilken:
For development packages, we cannot see what's changed between runs, so always rebuild the package in that case, and rely on e.g. CMake to figure out if anything needs to be rebuilt.

No change in behaviour if there are no untracked files.


TimoWilken:
Tested in a CI builder and working as far as I can tell.

PR #783, opened 2022-11-04
Time out when git takes too long

TimoWilken:
We want to avoid waiting for hours in CI if we can't update a git repository.

Add a `timeout=` argument to our `alibuild_helpers.cmd.*` functions and enforce it, at least where feasible, i.e. on Python 3.


TimoWilken:
Tested and working in Linux CI.

PR #785, opened 2022-11-21
Abort build when upload conflict on S3 is detected

TimoWilken:
If two aliBuild instances are building the same package at the same time and uploading it to S3, make sure that only one can upload at a time, and abort the other one.

There are probably ways to make aliBuild recover more intelligently from conflicts, but this commit only includes basic conflict detection.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -834,12 +834,6 @@ def doBuild(args, parser):
       if "obsolete_tarball" in spec:
         unlink(realpath(spec["obsolete_tarball"]))
         unlink(spec["obsolete_tarball"])
-      # We need to create 2 sets of links, once with the full requires,
-      # once with only direct dependencies, since that's required to
-      # register packages in Alien.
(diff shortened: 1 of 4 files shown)


TimoWilken:
I've set some non-critical nightly Jenkins builds to use this patched version of alibuild, namely DailyO2-ubuntu2004, DailyO2Dataflow-slc7 and DailyO2Dataflow-slc8.

TimoWilken:
One more change to do, discussed today: if there is a failure between uploading dist symlinks and the tarball, then the partial dist tree can block subsequent aliBuilds from uploading the same package. Moving dist symlink uploads to between the "main" symlink and the tarball in the store would solve this, as the "main" symlink reserves the revision number.

TimoWilken:
The latest commit should fix the problem mentioned in my previous comment. I've also rebased on master to fix merge conflicts.

I've set some non-essential Jenkins nightly builds (the o2-dataflow ones that are only used for CI caching) to use this branch in order to test the changes.

PR #786, opened 2022-12-24
wrapper script :: just add executable bits, do not overwrite system u…

adriansev:
…mask
@TimoWilken

### DIFF
--- a/alibuild_helpers/cmd.py
+++ b/alibuild_helpers/cmd.py
@@ -1,6 +1,5 @@
 import os
 import os.path
-import stat
 import sys
 import time
 from subprocess import Popen, PIPE, STDOUT
@@ -163,5 +162,5 @@ def install_wrapper_script(name, work_dir):
     exec env -u LD_LIBRARY_PATH -u DYLD_LIBRARY_PATH \\
          "$(which -a "$(basename "$0")" | grep -Fxv "$0" | head -1)" "$@"
     """))
-    os.fchmod(scriptf.fileno(), os.stat(scriptf.fileno()).st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH )  # make the wrapper script executable
+    os.fchmod(scriptf.fileno(), 0o755)  # make the wrapper script executable
   os.environ["PATH"] = script_dir + ":" + os.environ["PATH"]
--- a/alibuild_helpers/cmd.py
+++ b/alibuild_helpers/cmd.py
@@ -1,5 +1,6 @@
 import os
 import os.path
+import stat
 import sys
 import time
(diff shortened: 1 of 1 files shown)


adriansev:
actually this does not seem to work ... still looking for a solution for shared usage between multiple users that have a common group membership

adriansev:
i tried the simplest method (just use 775 and ignore any errors when chmod as most likely another alibuild already created there)

ktf:
I am not sure I understand the point. If it fails, it means that the script is still not executable and it will not work, no?

adriansev:
@ktf the context is that a group of users (as in users that pertain to a gid) use a common installation $ALIBUILD_WORK_DIR (which have a consistent common gid)
The current permissions set are 755, and the most simple solution would be to have 775, and the most elegant solution would be to check if overwriting is needed and if yes, after creation/overwriting just add execute bits for user and group (so there is not interference with the default umask)
I would be satisfied for the moment with just changing 755 to 775 ..
Would this be ok?
Thanks a lot!

TimoWilken:
Hi, sorry for sitting on this for a while.

My worry with the plain `0775` approach is that it would allow any user in the same group as me to overwrite a script, which aliBuild would automatically execute as me -- assuming that other user has execute access on the containing directories, all the way up to `/`. I guess in "normal" usage, this should be fine, since users should have their aliBuild workdir inside `$HOME`, which should have `0700` permissions by default, but still...

Presumably, in this multi-user environment, users should have a less restrictive `umask` than default, right? Perhaps it makes sense to set the file permissions to `0777 - umask` instead, since `umask` is `0022` by default. Or add just the executable bits to the existing default permissions, but it seems you tried that initially and ran into problems -- what was the issue there?

adriansev:
@TimoWilken So, since then the solution was for me to be the in-charged person for compilation of ALICE software on our cluster, so the problem is mostly solved :)))
The problem was was that the ALICE group have a single ALIBUILD_WORK_DIR location and software (that have default group ownership of the common `alice` group and now i set default acls g+rw)
The problem was that alibuild would try to overwrite `sw/wrapper-scripts/git` from https://github.com/alisw/alibuild/blob/master/alibuild_helpers/cmd.py#L157  each time and with 755. so, if a colleague tries to do another compilation the build will die with an error because the pre-existent `git` file with 755 permissions is not writable by group.
So .. the problem is solved for me because as root user i could set default acls, but maybe there are other groups that use a common cluster where a collaborative usage would reduce the redundant usage of space.. so, you decide if this is something worth spending time and work :)
Thanks a lot!

TimoWilken:
OK, thank you @adriansev. In that case, I'll close this PR since it's not currently mergeable, but I'll keep it in mind in future :)

PR #789, opened 2023-02-08
Use setuptools_scm to single-source an always-valid version number

TimoWilken:
The latest pip versions fail to install a source checkout if its version is set to `LAST_TAG`, which we do for all untagged commits.

Use setuptools_scm instead, so that we always have a valid, PEP440-compliant version number, even for commits in between tags.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,8 +17,6 @@ jobs:
         uses: actions/setup-python@v3
         with:
           python-version: 3.x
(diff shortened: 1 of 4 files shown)


TimoWilken:
The new coverage misses are in the error checking in `__init__.py` in case `aliBuild` is running from a local checkout and `setuptools_scm` is not installed. This seems unlikely to be worth testing using unittests. Merging.

PR #791, opened 2023-02-20
Remove `star()` function for experiment prefix

TimoWilken:
Instead of using a function to prefix `ali` to everything, just hardcode the prefix.

On Guix, Python packages are installed with a wrapper script, and the aliBuild executable is called `.aliBuild-real`, which breaks this and `aliBuild init` tries to clone `https://github.com/alisw/.alidist`.

### DIFF
--- a/aliBuild
+++ b/aliBuild
@@ -17,7 +17,7 @@ from alibuild_helpers.doctor import doDoctor
 from alibuild_helpers.deps import doDeps
 from alibuild_helpers.log import info, warning, debug, logger, error
 from alibuild_helpers.utilities import detectArch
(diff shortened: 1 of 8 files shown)


ktf:
we should check with the maintainers of the SHiP experiment. what is the rush to get this in?

TimoWilken:
Hi @ktf, sorry, I didn't realise anyone was still using this. I assumed this was only a holdover from the CMS-ALICE transition. Should I revert?

ktf:
let me check tomorrow and then we decide what to do.

TimoWilken:
OK, thanks!

PR #794, opened 2023-05-05
Protect against reusing partially-uploaded tarballs

TimoWilken:
Previously, if a package's symlinks were uploaded partially (e.g. not all dist symlinks uploaded, but the "main" symlink was present), and that package's hash matched what we need, then its revision number would be reused. This would then fail on upload, as aliBuild would detect a conflict with the partially-uploaded dist symlinks.

Relying on the fact that we upload the tarball last, after any symlinks, we can detect that this has happened by not reusing revision numbers for which we do not already have a tarball available. The `syncToLocal` function should download any tarball with the right hash, so if one is not available even though a symlink's hash matches, then we know it's probably from a partial upload and we must not reuse its revision number.

Draft since I've not tested this properly so far, I've only run some unittests locally.

@davidrohr this should fix the problem seen in the WeeklyO2 build last night. Though the specific partial upload of QC that was causing problems is now cleaned up -- I must've missed some files last night -- and a rebuild today went fine.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -709,10 +709,21 @@ def doBuild(args, parser):
         continue
       rev_hash, revision = match.groups()
 
-      if not (("local" in revision and rev_hash in spec["local_hashes"]) or
-              ("local" not in revision and rev_hash in spec["remote_hashes"])):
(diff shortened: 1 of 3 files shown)


TimoWilken:
I think this is not needed any more, since aliBuild should detect this another way already.

PR #798, opened 2023-08-08
Warn about uncommitted files in final summary

ktf:
Warn about uncommitted files in final summary

Too many times rebuilds have gone unnoticed because the per package
warning gets lost in the debug output. This tries to improve the issue
by adding a summary at the end with the list of directories containing
untracked files.


ktf:
Finally managed to fix the coverage of this. Merging.

PR #800, opened 2023-10-05
Generate etc/profile/init.sh early

TimoWilken:
This should help with debugging failed builds.

There is a change in behaviour, though: if we're unpacking a tarball, this means the init.sh from the tarball is used, not the newly-generated one. However, they should be the same.

Also, revamp and simplify the code that generates init.sh. Just use a single `cat<<\EOF` and generate init.sh's code directly from Python, instead of generating shell code that generates init.sh.

Also, make init.sh slightly more robust by quoting a few more things where applicable.

### DIFF
--- a/alibuild_helpers/build_template.sh
+++ b/alibuild_helpers/build_template.sh
@@ -13,7 +13,7 @@ function hash() { true; }
 export WORK_DIR="${WORK_DIR_OVERRIDE:-%(workDir)s}"
 
 # From our dependencies
-%(dependencies)s
+%(source_dependencies)s
 
 # Insert our own wrapper scripts into $PATH, patched to use the system OpenSSL,
 # instead of the one we build ourselves.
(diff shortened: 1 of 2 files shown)


alibuild_helpers/build_template.sh
```diff
@@ -84,6 +84,24 @@ if [[ $DEVEL_PREFIX ]]; then
...
+# from the tarball will replace the one we create here, but that should be
+# safe.
+mkdir -p "$INSTALLROOT/etc/profile.d"
+cat <<\EOF > "$INSTALLROOT/etc/profile.d/init.sh"
+# Source dependencies' init.sh.
+%(source_dependencies)s
+# Set up variables pointing to this package.
+%(package_vars)s
+# Environment variables set by this alidist recipe.
+%(environment)s
```
ktf, line 98: This might actually be the reason why we did not do this before. Environment variables created by a package should be effective only after the build. Does that change the behaviour?

ktf:
Can we have the revamp and the moving in two separate PRs? Way too many changes.

alibuild_helpers/build_template.sh
TimoWilken: No, it doesn't -- this block is inside the heredoc, so it only creates `init.sh`. It doesn't `source` it. Further up, only `%(source_dependencies)s` is executed.

The problem is that the order used to be "unpack tarball" -> "overwrite init.sh" (so the newly-generated init.sh wins); this changes it to "create init.sh" -> "unpack tarball, including init.sh, over the top" (i.e. tarball wins). But as I say in the comment, the files ought to be the same anyway I think...

PR #801, opened 2023-10-06
Move init.sh creation to before build and tarball unpacking

TimoWilken:
See also #800.

This changes a minimum of code, only moving the init.sh creation up in the build script.


ktf:
Thanks. Indeed environment is simply the creation of the environment file, no?

TimoWilken:
Yes, exactly!

TimoWilken:
Actually hang on, I think `.build-hash` should only be created after a successful build since aliBuild checks for it to see if a previous build has completed.

ktf:
yes, agreed.

TimoWilken:
Testing this, there's another problem: when a build fails, the `.envrc` points to `sw/<arch>/<package>/<version>/etc/profile.d/init.sh` (which would be correct if the build had succeeded), but it should instead point into `$INSTALLROOT` and only be changed to point to the former path after the build succeeds.

TimoWilken:
One more problem: if the build was done in a Docker container (like all the Linux CI builds are), then the paths will only be valid inside the container (i.e., they'll be under `/sw`, not `/build/nomad/.../ci/local/check/sw`).

PR #803, opened 2023-10-13
Abstract away SCM integration

ktf:
Abstract away SCM integration

The goal is to support different backends, like sapling.


TimoWilken:
I've commented on the combined PR at https://github.com/alisw/alibuild/pull/804.

TimoWilken (approved):

PR #806, opened 2023-10-17
Fix macOS detection when running standalone

ktf:
Fix macOS detection when running standalone


TimoWilken:
According to [the tox docs](https://tox.wiki/en/4.11.3/config.html#environment-variable-substitutions), you should be able to use `{env:VARIABLE_NAME}` in `commands` to substitute env vars.

TimoWilken:
Ah, you changed it already! :)

PR #808, opened 2023-10-18
Add support for rhel9

Barthelemy:


TimoWilken:
Hi Barth, thank you, but in the past we've always used slcX for all distros compatible with RHEL version X, e.g. slc8 for CentOS 8, Alma 8, RHEL 8 etc. If there's nothing special about RHEL9, I'd keep doing the same and just call it slc9.

Good point about adding it to the help message and `S3_SUPPORTED_ARCHS`, we should do that for rhel9 or slc9, whatever it ends up being called.

Barthelemy:
is it what you meant ?

TimoWilken:
I would also remove the unused "rhel" pattern (done in https://github.com/alisw/alibuild/pull/808/commits/4ccd30cc044030b89196d27e3b118f38f301cdac), but apart from that it looks perfect to me! I'll merge it once the checks are done.

TimoWilken (approved):
Thank you very much @Barthelemy!

awegrzyn:
Just FYI, currently `distro` package returns:
- ('Red Hat Enterprise Linux', '9.2', 'Plow') for RHEL9
- ('AlmaLinux', '9.2', 'Turquoise Kodkod') for Alma9

Therefore, with current implementation of `doDetectArch` we end up with: `rhel9_x86-64` and `slc9_x86-64` respectively.
I'd simply add `red hat enterprise linux` to https://github.com/alisw/alibuild/blob/master/alibuild_helpers/utilities.py#L153

TimoWilken:
Good catch, thanks @awegrzyn! I'll change it as you suggest.

PR #811, opened 2023-11-08
Fix SyntaxWarning on Python 3.12

TimoWilken:
Fix invalid escape sequences by escaping the backslash properly.

This stops Python showing a bunch of warnings on every aliBuild invocation.


TimoWilken:
This is a trivial change; merging.

PR #812, opened 2023-11-21
[O2-4412] Dynamically compute optional dependencies

awegrzyn:
Otherwise installation fails with:
```
 _MissingDynamic: `optional-dependencies` defined outside of `pyproject.toml` is ignored.
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = 'alibuild'
-dynamic = ['readme', 'version']
+dynamic = ['readme', 'version', 'optional-dependencies']
 description = 'ALICE Build Tool'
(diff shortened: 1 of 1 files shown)


awegrzyn:
@TimoWilken @ktf Could you have a look ASAP? This block me (and CI) from installing `aliBuild`.

TimoWilken:
Hi @awegrzyn, since we haven't supported running aliBuild with Python 2 for quite a while, I just deleted the key and released it as v1.15.3. Could you try with that?

awegrzyn:
> Hi @awegrzyn, since we haven't supported running aliBuild with Python 2 for quite a while, I just deleted the key and released it as v1.15.3. Could you try with that?

Yup, removing Python 2 support fixed the problem.

PR #814, opened 2023-11-25
Improve error reporting in aliBuild clean

TimoWilken:
1. Report the actual error when cleanup fails.
2. Keep going when one directory can't be deleted. This should stop one failing dir from completely blocking all cleanup in CI.
3. Use logging to show messages, to match the rest of aliBuild.
4. Fix decideClean docstring to match its code.
5. Adjust tests to match.


TimoWilken:
@ktf I've added tests to cover all the branches in doClean as well.

Could you merge this as well?

PR #815, opened 2023-11-28
Allow users to specify build comments

TimoWilken:
Write some more metadata for packages (augmenting and renaming the .full-dependencies file), including a user-supplied comment string, if any.

Users can supply comments using the new `--annotate PACKAGE=COMMENT` flag.

Currently, it is not checked whether the comment is actually applied. If the package is already installed, its comment is silently ignored.

### DIFF
--- a/alibuild_helpers/build_template.sh
+++ b/alibuild_helpers/build_template.sh
@@ -77,8 +77,8 @@ fi
 mkdir -p "$INSTALLROOT" "$BUILDROOT" "$BUILDDIR" "$WORK_DIR/INSTALLROOT/$PKGHASH/$PKGPATH"
 
 cd "$WORK_DIR/INSTALLROOT/$PKGHASH"
-cat <<\EOF > "$INSTALLROOT/.full-dependencies"
-%(dependenciesJSON)s
+cat > "$INSTALLROOT/.meta.json" <<\EOF
+%(provenance)s
 EOF
 
 # Add "source" command for dependencies to init.sh.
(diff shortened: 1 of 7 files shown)


alibuild_helpers/build_template.sh
```diff
@@ -80,8 +80,8 @@ fi
 mkdir -p "$INSTALLROOT" "$BUILDROOT" "$BUILDDIR" "$WORK_DIR/INSTALLROOT/$PKGHASH/$PKGPATH"
 
 cd "$WORK_DIR/INSTALLROOT/$PKGHASH"
-cat <<\EOF | tr \' \" >"$INSTALLROOT/.full-dependencies"
```
ktf, line 83: I think we do use `.full-dependencies` in a few places. I would simply add a new file.

TimoWilken: Do you know where? I can't find it being used anywhere in alibuild, ali-bot or alidist.

TimoWilken: Hi @awegrzyn, do you use the `.full-dependencies` file, e.g. in your RPM publisher?

awegrzyn: No, we do not use it atm, although if it also included build dependencies it would very useful for us.

ktf: OK, I suggest we remove it (as discussed, I think at some point we were using it), and add everything in the new file you have in this PR. Please make sure the full dependencies are there as well, so that @awegrzyn is happy.

TimoWilken: Done. @awegrzyn: the .meta.json file will have a `dependencies` key, which should include all the info you need (dependencies broken up by build/runtime and whether they're direct deps or transitive deps).

TimoWilken:
Rebased and fixed, CI is all green now.

PR #816, opened 2023-12-04
Regenerate init.sh after each build, in case it was clobbered

TimoWilken:
In particular, the AliEn-Runtime build overwrites init.sh. That makes subsequent builds fail, so regenerate init.sh after each build. (Re-)writing init.sh after the build matches the previous behaviour of alibuild.


TimoWilken:
Tested and working locally @ktf.

@pzhristov: This solves the issue with AliEn-Runtime you found.

PR #817, opened 2023-12-06
Fix missing environment variables when regenerating init.sh

TimoWilken:


TimoWilken:
@pzhristov This should solve the problem you saw with FLUKA_VMC.

I've tested this locally and it lets FLUKA_VMC compile properly for me, when it didn't before, so merging.

PR #818, opened 2023-12-07
Load the correct defaults file when loading packages

TimoWilken:
Previously, we would always load `defaults-release.sh`, even when another default was specified (such as `--defaults=o2`). Then, we'd override some of the keys of `defaults-release.sh` with those from the correct defaults file.

While we currently cover the most frequently-used keys in defaults (env, disable, overrides), this is obviously a massive footgun. It also means we must always have a `defaults-release.sh` file present, even if that default is never used.

This patch changes `getPackageList()` to load the correct underlying defaults file. It does not touch the override logic in `parseDefaults()`, which can probably be signficantly simplified after this patch is applied, but that's fiddly and would make this change less reviewable, so I'll leave it for another commit. This does mean that we'd load e.g. `defaults-o2.sh` and then override some of its keys with the same values loaded previously, but that shouldn't hurt.

Internally, all defaults are still called `defaults-release` to allow sharing packages between defaults (which I assume was the original motivation for doing this). There's not much point currently, since we set `env:` differently in all defaults, but we might want to use this in future.


ktf:
As discussed, we were indeed doing this to allow sharing packages between defaults.

PR #819, opened 2023-12-07
Generate init.sh in one place

TimoWilken:
This reduces the duplication in the init.sh generation code significantly.

Instead of generating shell commands that generate init.sh, scattered all over the place, generate the contents of init.sh in one place and just write them into the file when applicable.

Once created, load the generated file instead of generating the same shell code twice at different escaping levels.

All of this reduces the potential for errors a lot, since different places in the code don't need to be kept in sync when generating and re-generating the file.

Add tests that make sure the generated file contents are vaguely sensible, and that clobbering the file during the build (as the AliEn-Runtime package does) does not mean it is mangled during subsequent builds.

This PR requires #818 for the added tox test to work, so I've included that commit here. I'll rebase once that PR is merged.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -8,7 +8,7 @@ from alibuild_helpers.cmd import execute, getstatusoutput, DockerRunner, BASH, i
 from alibuild_helpers.utilities import prunePaths
 from alibuild_helpers.utilities import resolve_store_path
 from alibuild_helpers.utilities import format, parseDefaults, readDefaults
-from alibuild_helpers.utilities import getPackageList
+from alibuild_helpers.utilities import getPackageList, asList
 from alibuild_helpers.utilities import validateDefaults
 from alibuild_helpers.utilities import Hasher
 from alibuild_helpers.utilities import yamlDump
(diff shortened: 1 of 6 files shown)


TimoWilken:
Rebased and fixed another issue found by @pzhristov in AliRoot-OCDB. Ready to review/merge now for me.

ktf:
The feature itself is fine for me. My only worry is that the mangling at:

https://github.com/alisw/alibuild/pull/819/files#diff-44e897471db949fd0707d8ab0961b950f94569114948a039d69abaec52b63d7cL83

is not accounted correctly, however I guess we can simply merge this and try it for a bit before we tag it for production.

TimoWilken:
> The feature itself is fine for me. My only worry is that the mangling at:
> 
> https://github.com/alisw/alibuild/pull/819/files#diff-44e897471db949fd0707d8ab0961b950f94569114948a039d69abaec52b63d7cL83
> 
> is not accounted correctly, however I guess we can simply merge this and try it for a bit before we tag it for production.

Ah, I changed that to a proper `json.dumps` call, since `json` has been built-in to the Python standard library for a long time, and the original way of stringifying a Python dict and replacing the quotes seemed really fragile to me.

With https://github.com/alisw/alibuild/pull/815, that'll be replaced anyway though...

PR #822, opened 2023-12-13
Avoid finding false-positive symlinks and printing warnings for them

TimoWilken:
If users build e.g. O2Physics first with a branch called "x-y", and then later with a branch called "x", then aliBuild would print warnings about not being able to parse symlinks.

This is caused by an overly-broad glob when listing symlinks.

Replace this glob with a regex modelled after the regex that parses symlink targets, to avoid these false positives.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -32,6 +32,7 @@ except ImportError:
   from pipes import quote  # Python 2.7
 
 import concurrent.futures
+import errno
 import importlib
 import json
 import socket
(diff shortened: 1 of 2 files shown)


TimoWilken:
@vkucera This should get rid of the extraneous warnings you saw!

vkucera:
> @vkucera This should get rid of the extraneous warnings you saw!

Thanks @TimoWilken !

ktf:
Could you cover also the case the symlink returns NOENT?

TimoWilken:
Thinking about this a bit more, it's probably better to handle all cases where we can't access the symlink dir by continuing as if it was empty. That fixes the coverage as well.

PR #823, opened 2023-12-19
Fix force_rebuild setting when specified by recipe

TimoWilken:
Previously, alibuild would always overwrite the `force_rebuild:` YAML key with the value of the `--force-rebuild` command-line flag, so the former would have no effect.


TimoWilken:
Here, I don't understand why the overall coverage decreased. The only decrease that codecov sees is [in the aliBuild script](https://app.codecov.io/gh/alisw/alibuild/pull/823/indirect-changes), but this is stuff we've never tested, as far as I can tell (the `--profile` option and interrupting aliBuild while it runs).

PR #827, opened 2024-01-31
Install pigz by default on Ubuntu

TimoWilken:
Recommend the installation of pigz on Ubuntu as a dependency of alibuild. This means pigz will be installed by default, unless the user explicitly runs `apt install --no-install-recommends python3-alibuild`.


ktf:
Did you push the right commit?

TimoWilken:
Yes, I only want to change `debian/control`. Do you want me to remove the change in the description line?

PR #830, opened 2024-02-26
Make `aliBuild build` much more efficient

TimoWilken:
* Only fetch git repos for packages that we'll actually compile. If we download a tarball, we don't need the repo. This saves downloading e.g. the massive Clang repo.

* If a package is already installed, don't (re-)fetch a tarball for it. This stops us from constantly redownloading useless tarballs after `aliBuild clean --aggressive-cleanup`.

* Make it more obvious to the user what is happening, specifically whether a package is being compiled or downloaded and unpacked. Show git-clone and download progress. In general, prefer showing progress percentages instead of showing lots of output in non-debug mode.

* Do as much work as possible before we start to compile the first package, so that (a) more info is available to plugins, and (b) we can print more helpful messages. Drop the weird control flow where we'd try to build everything twice, just to detect that the package was already built on the second try. This makes log messages much less confusing.

* Update tests to match.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -1,160 +1,109 @@
-from os.path import abspath, exists, basename, dirname, join, realpath
-from os import makedirs, unlink, readlink, rmdir
-from alibuild_helpers import __version__
-from alibuild_helpers.analytics import report_event
-from alibuild_helpers.log import debug, error, info, banner, warning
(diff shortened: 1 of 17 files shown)


TimoWilken:
Closing this in favour of https://github.com/alisw/alibuild/pull/832 and future PRs building on top of that.

PR #832, opened 2024-02-28
Preparatory cleanups for aliBuild refactoring

TimoWilken:
This contains most of the cleanups and preparation for https://github.com/alisw/alibuild/pull/830, so that I can send future PRs with just the "meat" of the refactoring changes.

Please see the individual commits below for changes and reasoning.

I hope that none of them are particularly controversial.

I've encapsulated changes that belong together into their own commits, to make it easier to review the PR. It's probably easier to look at each commit on its own, rather than the combined diff.


TimoWilken:
All green now, ready for review @ktf!

aliBuild
```diff
@@ -81,7 +81,8 @@ def doMain(args, parser):
     exit(0)
 
   if args.action == "build":
-    sys.exit(doBuild(args, parser))
+    doBuild(args, parser)
```
ktf, line 84: this changes the behavior, no? doesn't doBuild return the exit code?

PR #834, opened 2024-03-05
UI improvements and tarballs fix

TimoWilken:
See individual commits.

* Only download tarballs if we're going to use them (fixes https://its.cern.ch/jira/browse/O2-3882)
* Show download progress interactively (if possible) using `ProgressPrint`
* Show progress of cloning "mirror" repositories, replacing "spammy" per-repo output with one `ProgressPrint` (in non-debug mode)

Also contains one cleanup/refactoring commit (https://github.com/alisw/alibuild/commit/4236ab3044778141605f6e493ad275c46963daf5) -- let me know if you want a separate PR for that one.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -55,7 +55,7 @@ def readHashFile(fn):
     return "0"
 
 
-def update_git_repos(args, specs, buildOrder, develPkgs):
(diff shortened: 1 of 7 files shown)


TimoWilken:
Squashed and rebased @ktf!

PR #835, opened 2024-03-05
Clone SOURCES repos outside container

TimoWilken:
This lets us always use speed-up options like `--filter`, even when we're building inside a container.

We can also easily start using sapling for this in future by implementing the new SCM methods.

This also helps with some annoying authentication-related issues, if running inside `--docker`, such as in CI.

Fixes <https://its.cern.ch/jira/browse/O2-2439>.

When this is merged, we can also get rid of the `GIT_*` environment variables used in CI, and stop installing an especially-new Git version in our containers (since that was needed to understand the `GIT_*` variables).

This PR is based on https://github.com/alisw/alibuild/pull/834; once that is merged, I can rebase to get rid of the common commits.

### DIFF
--- a/alibuild_helpers/build.py
+++ b/alibuild_helpers/build.py
@@ -65,27 +65,19 @@ def update_git_repos(args, specs, buildOrder):
     def update_repo(package, git_prompt):
         specs[package]["scm"] = Git()
         if specs[package]["is_devel_pkg"]:
-          localCheckout = os.path.join(os.getcwd(), specs[package]["package"])
-          if exists("%s/.sl" % localCheckout):
-            specs[package]["scm"] = Sapling()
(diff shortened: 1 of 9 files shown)


TimoWilken:
Rebased on https://github.com/alisw/alibuild/pull/834. Only the last two commits are actually new!

ktf:
As discussed, let's have a prerelease and try it out in one of the non-production CIs.

TimoWilken:
I've released [v1.17.0rc1](https://github.com/alisw/alibuild/releases/tag/v1.17.0rc1) and set the Ubuntu alidist CI to use it in https://github.com/alisw/ali-bot/commit/56ec67d533218e022fa406b0b755658d2b00847e!

PR #837, opened 2024-03-13
Fix source checkout with bare commit hash

TimoWilken:


pzhristov:
I confirm this PR solves the problem with UUID.

TimoWilken:
Perfect, thank you @pzhristov!

PR #846, opened 2024-06-05
Consider all git repositories managed by aliBuild as safe.

ktf:
Consider all git repositories managed by aliBuild as safe.


olantwin:
@ktf Are you planning to make a release for this?

If not, we'll just pin the latest commit from master for the time being.

ktf:
I have just done a release.

olantwin:
Thanks!

PR #849, opened 2024-06-18
Include clone_speedup_options in git fetch

singiamtel:
This change is not only performance related, it should also fix the missing blob object errors. Fetching with --filter=tree:0 should always be more resilient to missing objects upstream

```
sh-5.1$ git fetch -f https://github.com/alisw/alibuild-recipe-tools
+refs/tags/*:refs/tags/* +refs/heads/*:refs/heads/*
remote: Enumerating objects: 8, done.
remote: Counting objects: 100% (8/8), done.
remote: Compressing objects: 100% (4/4), done.
remote: Total 6 (delta 2), reused 6 (delta 2), pack-reused 0
Unpacking objects: 100% (6/6), 806 bytes | 806.00 KiB/s, done.
fatal: missing blob object '6f314120ba0359d1f73ad7d1d4a7f171fb36e8ac'
error: https://github.com/alisw/alibuild-recipe-tools did not send all necessary objects
```

### DIFF
--- a/alibuild_helpers/git.py
+++ b/alibuild_helpers/git.py
@@ -8,13 +8,8 @@ GIT_COMMAND_TIMEOUT_SEC = 120
 """How many seconds to let any git command execute before being terminated."""
 
 
-# Don't recalculate this every time we need it.
-_clone_speedup_cache = None
 def clone_speedup_options():
   """Return a list of options supported by the system git which speed up cloning."""
-  if _clone_speedup_cache is not None:
-    return _clone_speedup_cache
-
   for filter_option in ("--filter=tree:0", "--filter=blob:none"):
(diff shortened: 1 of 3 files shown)


alibuild_helpers/git.py
```diff
@@ -8,8 +8,13 @@
 """How many seconds to let any git command execute before being terminated."""
 
 
+# Don't recalculate this every time we need it.
+_clone_speedup_cache = None
 def clone_speedup_options():
   """Return a list of options supported by the system git which speed up cloning."""
+  if _clone_speedup_cache is not None:
```
ktf, line 15: Where do you actually set this?

ktf:
Can you adapt the tests? Also, I would make the caching a separate PR, just to make sure something does not depend on the command being executed every time (e.g. inside and outside a container).

PR #851, opened 2024-06-19
Migrate to mkdocs

singiamtel:
Reopening this PR

---

Currently running on http://ali-ci-desktop.cern.ch/

Migrate from Jekyll to [mkdocs](https://www.mkdocs.org/). The main feature we want from mkdocs is the search bar

I've tried to keep all the old links working, but there may be some that are broken. If you find any, please let me know.

The style is from https://config.docs.cern.ch/

- [X] Migrate from Jekyll to mkdocs
- [X] Setup GH actions auto deployment
- [X] Fix docs tests

To try it, install the dependencies and run ``mkdocs serve``

```bash
# dependencies
pip install mkdocs mkdocs-material mkdocs-markdownextradata-plugin mkdocs-redirects
```


singiamtel:
I've been trying other themes but they don't look great. We can change the colorscheme/inject CSS to this one if you don't like something about it (https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#configuration)

PR #860, opened 2024-07-17
Support CVMFS based remote store

ktf:
Support CVMFS based remote store

This introduces experimental support for reusing packages which
were built with aliBuild and were deployed on CVMFS (or similar shared
filesystem).

It works by locating the package with the correct hash, and creating a
tarball with a symlink to it. For this reasons, this cannot be used
to upload tarballs.


ktf:
Merging to simplify development. It's must be requested by hand in any case.

PR #871, opened 2024-09-27
Remove most Python 2 references

singiamtel:
Probably not ready to be merged, making PR to run the tests.

Also fixing some minor issues with the inheritance in ``class Git(SCM)``

### DIFF
--- a/alibuild_helpers/cmd.py
+++ b/alibuild_helpers/cmd.py
@@ -1,31 +1,15 @@
 import os
 import os.path
-import sys
 import time
 from subprocess import Popen, PIPE, STDOUT
 from textwrap import dedent
-try:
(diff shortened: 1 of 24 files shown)


alibuild_helpers/cmd.py
```diff
@@ -1,31 +1,15 @@
...
-if sys.version_info[0] >= 3:
-  basestring = str
-  unicode = None
-
-
 def is_string(s):
-  if sys.version_info[0] >= 3:
-    return isinstance(s, str)
-  return isinstance(s, basestring)
+  return isinstance(s, str)
```
ktf, line 12: maybe we can simplify this further?

ktf:
One thing at the time, this is already quite a lot of changes to review, albeit trivial.

PR #875, opened 2024-10-15
Allow definying classes of equivalence for system tools

ktf:


ktf:
@singiamtel this will simplify the prefer_system replacements we discussed today.

PR #882, opened 2024-11-12
Improve AliBuild typing and linter strictness

singiamtel:
I'm currently trying to type the codebase as much as possible, and use a stronger mypy filter.

~~Not ready to merge yet, just want to get the tests running~~


alibuild_helpers/sync.py
```diff
@@ -67,8 +67,9 @@ def getRetry(self, url, dest=None, returnResult=False, log=True, session=None, p
           reportTime = time.time()
           result = []
 
+          destFp = None
           try:
-            destFp = open(dest+".tmp", "wb") if dest else None
+            if dest: destFp = open(dest+".tmp", "wb")
```
ktf, line 72: Not sure this is an improvement.

ktf: maybe we should split the opening of the file from the rest of the try catch...

ktf: I would live this particular change out and continue with the rest of the PR, since to that I have no particular objections.

aliBuild
```diff
@@ -111,8 +111,8 @@ if __name__ == "__main__":
   else:
     os.environ["ALIBUILD_ANALYTICS_USER_UUID"] = open(expanduser("~/.config/alibuild/analytics-uuid")).read().strip()
   try:
-    profiler = "--profile" in sys.argv
```
ktf, line 114: please avoid renaming variables just for the sake of renaming them.

singiamtel:
Closing this for now, will continue after the rebase

Issue #887, opened 2024-11-19
A build results in formatting error


mborisyak:
Dear alibuild developers,

I was using `aliBuild` for compiling [FairShip](https://github.com/ShipSoft/shipdist) and encountered the following error:
```bash
$ aliBuild build FairShip --config-dir shipdist/ --defaults release --always-prefer-system
Traceback (most recent call last):
  File "/home/max/opt/pyenv/versions/py3/bin/aliBuild", line 130, in <module>
    doMain(args, parser)
  File "/home/max/opt/pyenv/versions/py3/bin/aliBuild", line 83, in doMain
    doBuild(args, parser)
  File "/home/max/opt/pyenv/versions/3.10.4/envs/py3/lib/python3.10/site-packages/alibuild_helpers/build.py", line 490, in doBuild
    getPackageList(packages                = packages,
  File "/home/max/opt/pyenv/versions/3.10.4/envs/py3/lib/python3.10/site-packages/alibuild_helpers/utilities.py", line 469, in getPackageList
    requested_version = resolve_version(spec, defaults, "unavailable", "unavailable")
  File "/home/max/opt/pyenv/versions/3.10.4/envs/py3/lib/python3.10/site-packages/alibuild_helpers/utilities.py", line 126, in resolve_version
    return spec["version"] % {
TypeError: unsupported operand type(s) for %: 'float' and 'dict'
```
The error was provoked by [the following dependency](https://github.com/ShipSoft/shipdist/blob/main/termcap.sh#L2):
```
package: termcap
version: 1.0
```
It seems like the version was parsed as float.

Forcing `str` in [alibuild_helpers/utilities.py:126](https://github.com/alisw/alibuild/blob/master/alibuild_helpers/utilities.py#L126) does solve the problem:
```python
return str(spec['version']) % {
  ...
}
```
however, I'm not sure if it is the right way to fix the issue.

Sincerely,
Maxim.

mborisyak:
Thank you!
An alternative would be to raise an error, but you are right --- converting the version into a string is easier and more intuitive.

PR #888, opened 2024-11-19
Ensure spec["version"] is a string before trying to expand it

singiamtel:
Some valid versions defined in the recipe can be parsed as floats
instead (e.g. 1.0)

Fix by @mborisyak

Closes https://github.com/alisw/alibuild/issues/887

Further discussion here: https://github.com/ShipSoft/shipdist/issues/94

### DIFF
--- a/alibuild_helpers/utilities.py
+++ b/alibuild_helpers/utilities.py
@@ -123,7 +123,7 @@ def resolve_version(spec, defaults, branch_basename, branch_stream):
   defaults_upper = defaults != "release" and "_" + defaults.upper().replace("-", "_") or ""
   commit_hash = spec.get("commit_hash", "hash_unknown")
   tag = str(spec.get("tag", "tag_unknown"))
-  return spec["version"] % {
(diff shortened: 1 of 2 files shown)


alibuild_helpers/utilities.py
```diff
@@ -123,7 +123,7 @@ def resolve_version(spec, defaults, branch_basename, branch_stream):
   defaults_upper = defaults != "release" and "_" + defaults.upper().replace("-", "_") or ""
   commit_hash = spec.get("commit_hash", "hash_unknown")
   tag = str(spec.get("tag", "tag_unknown"))
-  return spec["version"] % {
```
ktf, line 126: I am not sure this will work correctly. For example the float `1.00` which might be intended as the version `"1.00"` will happily convert to `"1.0"`. If you really want to fix this, it needs to be done at the level of the parser which should always consider "version" to be followed by a string.

ktf (changes requested):
This will not work, see inline comment.

PR #893, opened 2024-12-13
Warn on implicit no remote store

singiamtel:
Make it more clear when we're recompiling everything ourselves

I'm adding it here instead of in `alibuild_helpers/args.py` because the logger is not initialized yet during the arg parsing

I'm honestly not very convinced about how this change turned out, if there's any suggestion on how to improve it, please let me know

### DIFF
--- a/tests/test_build.py
+++ b/tests/test_build.py
@@ -264,6 +264,7 @@ class BuildTestCase(unittest.TestCase):
         args = Namespace(
             remoteStore="",
             writeStore="",
+            no_remote_store=True,
             referenceSources="/sw/MIRROR",
             docker=False,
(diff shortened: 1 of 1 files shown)


alibuild_helpers/build.py
```diff
@@ -451,6 +452,11 @@ def doBuild(args, parser):
              'Maybe you need to "cd" to the right directory or '
              'you forgot to run "aliBuild init"?' % args.configDir)
 
+  if (not args.no_remote_store and
```
ktf, line 455: The change itself is fine for me. What happens on platforms where we prefer the system tools, like on macOS?

Could you add a test which exercises this part of the code?

PR #901, opened 2025-01-27
Allowing enabling dependencies via defaults

ktf:


ktf:
Coverage report is spurious. We should probably exclude documentation from the coverage information.

PR #906, opened 2025-02-19
Remove pkg_resources

singiamtel:
It's deprecated and removed since Python 3.12


ktf:
This will break the current way of packaging things with homebrew.

ktf:
(i.e., we need a way to mimick pkg_resources)

ktf (changes requested):
Maybe we should switch to importlib.

PR #919, opened 2025-04-14
Prefer system check which result in a new recipe should be treated as normal packages

ktf:
In general when are remote store is provided, we disable system packages. However, this
should not be the case if the system package results in a new recipe, because in that case
we consider the result of that recipe to be the package.


ktf:
@davidrohr this is what we disucssed.
@singiamtel could you cross check the logic? The idea is that the system package replacements override the restriction on system packages when a remote store is specified (because in general we do not want to uploaded packages depending on the presence of system packages).

ktf:
@singiamtel can we have a release with this?

singiamtel:
LGTM. New release ongoing

PR #920, opened 2025-04-15
Only upload blessed system packages

ktf:


ktf:
@singiamtel this should prevent we upload system packages we do not want to upload.

@davidrohr you need to add `allow_system_package_upload: True` to your `gpu-system.sh` recipe for it to work correctly.

PR #922, opened 2025-04-15
Bump ubuntu builder

singiamtel:
20.04 is EOL as of today


singiamtel:
@ktf should we drop the test for 3.6? It's no longer supported by Github actions https://github.com/actions/runner-images/issues/11101

PR #924, opened 2025-05-07
Allow building in a specified directory

ktf:
This is for the case where workdir is on a slow filesystem, e.g.
NFS.

Notice we keep devel packages in the old (possibly slow) place,
so that we can do incremental builds (in case the "fast" path has some garbage collection like /tmp).


ktf:
@davidrohr @ehellbar this is what I was mentioning today in the meeting. It allows you to do:

```
 ALIBUILD_BUILD_WORK_DIR=/tmp/eulisse python3 ~/src/alibuild/aliBuild --config ~/src/alidist --defaults o2 build O2PDPSuite
```

and have any required building which is not for a devel package in /tmp.

PR #927, opened 2025-05-21
Add support for tracking external environment

ktf:
A recipe can now provide a dict in `track_env` where the
key is the name of an environment variable and the value
is a command which is executed to fill such environment variable.

This can be used to track system properties or environment variables
without having to resort to a system package.


ktf:
Does not change anything if track_env is not in the yaml, so we should be fine.

PR #928, opened 2025-05-22
Change entrypoint of aliBuild

singiamtel:
This PR does the following:

# Change entrypoint of aliBuild

pyproject.toml has a `[project.scripts]` section that defines the entry point for the aliBuild command. This let's us get rid of the helper scripts we used to have.

We could keep `/aliBuild` as an alias to `/alibuild_helpers/main.py` in order not to break muscle memory, I have no preference. 

Otherwise you'd need to run `pip install -e .`, and then `aliBuild` should point to the local one.

# Update release workflow for pyproject.toml

We still had the old `setup.py` in the release workflow. This is now replaced by a call to `build` which will use the `pyproject.toml` file to build the package.

### DIFF
--- a/alibuild_helpers/main.py
+++ b/alibuild_helpers/main.py
@@ -83,7 +83,11 @@ def doMain(args, parser):
     sys.exit(0)
 
 
-if __name__ == "__main__":
+def main(action = None):
+  if action:
+    sys.argv[0] = "aliBuild"
+    sys.argv.insert(1, action)
+
   args, parser = doParseArgs()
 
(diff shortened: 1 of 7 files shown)


alibuild_helpers/main.py
ktf, line 1: Why is this a new file? Also, this means that one cannot clone alibuild anymore and execute, say aliDoctor. Am I wrong? If yes, we should fix all the associated documentation as well.

ktf (approved):

PR #930, opened 2025-05-28
Run prefer_system_check in a temporary dir

singiamtel:
This way we don't leave traces in case the check happened to create
temporary files

CC @ktf @davidrohr

### DIFF
--- a/alibuild_helpers/cmd.py
+++ b/alibuild_helpers/cmd.py
@@ -37,9 +37,9 @@ def getoutput(command, timeout=None):
   return decode_with_fallback(stdout)
 
 
-def getstatusoutput(command, timeout=None):
+def getstatusoutput(command, timeout=None, cwd=None):
   """Run command and return its return code and output (stdout and stderr)."""
-  proc = Popen(command, shell=isinstance(command, str), stdout=PIPE, stderr=STDOUT)
+  proc = Popen(command, shell=isinstance(command, str), stdout=PIPE, stderr=STDOUT, cwd=cwd)
   try:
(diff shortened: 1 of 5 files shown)


alibuild_helpers/cmd.py
```diff
@@ -91,11 +91,12 @@ def __enter__(self):
...
       self._container = getoutput(cmd).strip()
 
-    def getstatusoutput_docker(cmd):
+    def getstatusoutput_docker(cmd, cwd=None):
       if self._container is None:
-        return getstatusoutput("{} -c {}".format(BASH, quote(cmd)))
+        return getstatusoutput("{} -c {}".format(BASH, quote(cmd)), cwd=cwd)
       return getstatusoutput("docker container exec {} bash -c {}"
-                             .format(quote(self._container), quote(cmd)))
+                             .format(quote(self._container), quote(cmd)),
```
ktf, line 98: Does this actually work? It will change the cwd of where the container executes stuff, not the one of the actual command, no?

singiamtel: It doesn't make much sense to change it for docker in this case, but in any case the docker container will be disposed of as soon as the checks are done. 

This change mostly matters to not leave random stuff on the host

ktf: Is this addressed?

ktf:
Sounds good. There is a testdist you can use for this recipe. Maybe you could add it there and add a test? Thanks.

alibuild_helpers/cmd.py
ktf: Can you try?

PR #934, opened 2025-06-04
Add location of alidist as ALIBUILD_CONFIG_DIR variable

ktf:

PR #935, opened 2025-06-05
Allow for extra variables, volumes in the docker runner

ktf:


singiamtel:
@ktf seems to work now. I'd like to standarize how we pass variables to each environment eventually (and update the documentation), I find it overly complex now.

@davidrohr Can you confirm whether this does what you want?

davidrohr:
@singiamtel : It works, but it prints only `alidist`, but I need the full absolute path, since I am running in a temporary dir.

davidrohr:
With your new fix it works

alibuild_helpers/build.py
```diff
@@ -501,7 +501,7 @@ def doBuild(args, parser):
 
   install_wrapper_script("git", workDir)
 
-  with DockerRunner(args.dockerImage, args.docker_extra_args) as getstatusoutput_docker:
+  with DockerRunner(args.dockerImage, args.docker_extra_args, extra_env={"ALIBUILD_CONFIG_DIR": os.path.abspath(args.configDir)}) as getstatusoutput_docker:
```
ktf, line 504: I guess we need to mount alidist as a volume?

alibuild_helpers/deps.py
```diff
@@ -16,7 +16,10 @@ def doDeps(args, parser):
   specs = {}
   defaultsReader = lambda: readDefaults(args.configDir, args.defaults, parser.error, args.architecture)
   (err, overrides, taps) = parseDefaults(args.disable, defaultsReader, debug)
-  with DockerRunner(args.dockerImage, args.docker_extra_args) as getstatusoutput_docker:
+  with DockerRunner(args.dockerImage, args.docker_extra_args, extra_env={"ALIDIST_PATH": "/alidist" if args.docker else args.configDir}, extra_volumes=[f"{os.path.abspath(args.configDir)}:/alidist:ro"] if args.docker else []) as getstatusoutput_docker:
```
ktf, line 19: shouldn't this be `os.path.absPath(args.configDir)` for ALIDIST_PATH as well?

tests/test_cmd.py
```diff
@@ -38,6 +38,36 @@ def test_DockerRunner(self, mock_getstatusoutput, mock_getoutput):
             mock_getstatusoutput.reset_mock()
         mock_getstatusoutput.assert_not_called()
 
+    @mock.patch("alibuild_helpers.cmd.getoutput")
+    @mock.patch("alibuild_helpers.cmd.getstatusoutput")
+    def test_DockerRunner_with_env_vars(self, mock_getstatusoutput, mock_getoutput):
+        """Test that environment variables are properly injected into docker exec commands."""
```
ktf, line 44: Please, make this an actual comment. Why does it need to be a function comment?

ktf:
Thank you for following it up and sorry for the delay. Could you please have a look at the comments? I also do not understand why the code coverage went down. Do you? Please also do not add any cosmetic refactoring (e.g. os.remove vs remove) unless it's part of the feature.

alibuild_helpers/build.py
```diff
@@ -501,7 +501,7 @@ def doBuild(args, parser):
 
   install_wrapper_script("git", workDir)
 
-  with DockerRunner(args.dockerImage, args.docker_extra_args) as getstatusoutput_docker:
+  with DockerRunner(args.dockerImage, args.docker_extra_args, extra_env={"ALIBUILD_CONFIG_DIR": os.path.abspath(args.configDir), "ALIDIST_PATH": "/alidist" if args.docker else args.configDir}, extra_volumes=[f"{os.path.abspath(args.configDir)}:/alidist:ro"] if args.docker else []) as getstatusoutput_docker:
```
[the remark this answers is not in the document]
singiamtel: I didn't realize that we had ALIBUILD_CONFIG_DIR already from #934. Indeed they seem to mean the same, which one would you rather keep?

singiamtel:
> @singiamtel : No, unfortunately it still fails due to the path being relative, `$ALIBUILD_CONFIG_DIR=alidist`.

My bad, I had forgotten one. Should be fixed now

davidrohr:
@singiamtel : Thx, now it works!

PR #937, opened 2025-06-16
Replace %(key)s in environment variables during prefer_system_check

singiamtel:
@davidrohr does this work for you? I had to inject the env variables into the
recipe too, apparently they weren't before

### DIFF
--- a/alibuild_helpers/utilities.py
+++ b/alibuild_helpers/utilities.py
@@ -8,6 +8,7 @@ import sys
 import os
 import re
 import platform
+import copy
 
 from datetime import datetime
 from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


alibuild_helpers/utilities.py
```diff
@@ -582,7 +584,20 @@ def getPackageList(packages, specs, configDir, preferSystem, noSystem,
             spec = replacement
             # Allows generalising the version based on the actual key provided
             spec["version"] = spec["version"].replace("%(key)s", key)
+            # Also replace %(key)s in environment variables
+            if "env" in spec:
+              for env_key, env_value in spec["env"].items():
+                if isinstance(env_value, str):
+                  spec["env"][env_key] = env_value.replace("%(key)s", key)
             recipe = replacement.get("recipe", "")
+            # Inject environment variables into the recipe since they're not available during execution
```
ktf, line 593: This changes the current behaviour, no? env is not used for the current package, but only for it's dependencies.

PR #938, opened 2025-06-19
Fix Rocky Linux detection

sfayer:
Hi,

We've found that on Rocky9, the platform detection fails. This seems to be caused by the platform name containing a space, so I've adjusted the expected value to match:
```
>>> distro.linux_distribution()
('Rocky Linux', '9.6', 'Blue Onyx')
```

Regards,
Simon


ktf:
Please add coverage.

sfayer:
Hi,

Please find the coverage tests added. I don't think "rockylinux" without a space has ever been used, but I can add it back into the list if you'd like?

Regards,
Simon

PR #947, opened 2025-07-04
Load CLI environment variables in prefer_system_check

singiamtel:


tests/test_deps.py
```diff
@@ -64,7 +64,8 @@ def depsOpen(fn, mode):
                          outdot="/tmp/out.dot",
                          outgraph="/tmp/outgraph.pdf",
                          package="AliRoot",
-                         defaults="release")
+                         defaults="release",
+                         environment=[])
```
ktf, line 68: Why a list here and not below?

PR #951, opened 2025-07-15
Fix BUILD folder symlinks

singiamtel:
Oversight from #924


alibuild_helpers/build_template.sh
```diff
@@ -103,9 +103,9 @@ unset DYLD_LIBRARY_PATH
 EOF
 
 cd "$BUILDROOT"
-ln -snf $PKGHASH "${BUILDROOT}-latest"
```
ktf, line 106: I do not understand. Above we do:

```
export BUILDROOT="$ALIBUILD_BUILD_WORK_DIR/BUILD/$PKGHASH"
```

why the explicit expansion? Also, why dropping the quotes?

singiamtel: The quotes are an oversight, just fixed them.

I took the structure from the old code, the new symlinks were called hash-latest rather than name-latest, and therefore were not cleaned up after

Before #924 
```bash
export BUILDROOT="$WORK_DIR/BUILD/$PKGHASH"
[...]
ln -snf $PKGHASH $WORK_DIR/BUILD/$PKGNAME-latest
 if [[ $DEVEL_PREFIX ]]; then
  ln -snf $PKGHASH $WORK_DIR/BUILD/$PKGNAME-latest-$DEVEL_PREFIX
 fi
 ```
 
 After #924 
 ```bash
 export BUILDROOT="$ALIBUILD_BUILD_WORK_DIR/BUILD/$PKGHASH"
 [...]
ln -snf $PKGHASH "${BUILDROOT}-latest"
 if [[ $DEVEL_PREFIX ]]; then
  ln -snf $PKGHASH "${BUILDROOT}-latest-$DEVEL_PREFIX"
 fi
 ```
 
 After this PR
 ```bash
 export BUILDROOT="$ALIBUILD_BUILD_WORK_DIR/BUILD/$PKGHASH"
[...]
ln -snf "$PKGHASH" "$ALIBUILD_BUILD_WORK_DIR/BUILD/$PKGNAME-latest"
 if [[ $DEVEL_PREFIX ]]; then
  ln -snf "$PKGHASH" "$ALIBUILD_BUILD_WORK_DIR/BUILD/$PKGNAME-latest-$DEVEL_PREFIX"
 fi
 ```

ktf: Right.

PR #955, opened 2025-07-22
Fix multiline env variables

singiamtel:
Regression in #947

CC @ktf


ktf:
Please also add the unit test for the case which we handle here.

PR #958, opened 2025-08-06
Use $HOME variable over ~

singiamtel:
Otherwise some CI machines can get stuck with a
/workdir/~/.config/alibuild/analytics-uuid file


ktf:
I think this can only happen when HOME is not set, so it will then try to write in a place where you might (or might not) have permissions to. IMHO a better option is to check for $HOME to be there and be writeable.

ktf:
`HOME= bash -c "ls ~"` to verify

PR #971, opened 2025-09-25
[O2-6138] Parallelize symlinks

singiamtel:
Parallelize symlinks upload at the end of the build

I can't test it without actually uploading things to S3, I'd suggest we can run some of the less important dailies with this branch for a while and see if it brings any issue/regression


ktf:
Ok, thanks.

